PR #64, opened 2015-01-28
Externalised padding for .block-inner

LukaszGrela:
During the course styling I've found that moving the padding of `.block-inner` to `variables.less` is more convenient.


taylortom:
+1

oliverfoster:
+1

hrajotia:
+1

PR #65, opened 2015-02-11
Accessibility updated to make accessible

oliverfoster:


moloko:
+1 - though probably be good for someone with better CSS skills than I have to review as well

dancgray:
+1

templates/navigation.hbs
```diff
@@ -1,6 +1,6 @@
 {{! Maintainers - Daryl Hedley}}
 {{! All links in the nav bar should have an attritbue of data-event, navigationView uses this to fire a global event}}
-<div class="navigation-inner clearfix" role="navigation">
-    <a href="#" class="navigation-back-button icon icon-controls-small-left" data-event="backButton" role="button" aria-label="{{previous}}"></a>
-    <a href="#" class="navigation-drawer-toggle-button icon icon-list" data-event="toggleDrawer" role="button" aria-label="{{navigationDrawer}}"></a>
+<div class="navigation-inner clearfix" role="navigation" aria-label="{{_accessibility._ariaLabels.navigation}}">
```
dancgray: Should this be _globals._accessibility._ariaLabels.navigation?

lc-petranussdorfer:
+1

templates/navigation.hbs
oliverfoster: https://github.com/adaptlearning/adapt_framework/blob/develop/src/core/js/views/navigationView.js#L34

^ This line says not. The _globals variable becomes the model for the navigation view. But we could certainly change it to make them all uniform... I'll do separate issues and PR's for both the theme template and the core view.

PR #68, opened 2015-03-12
Accessibility - Fixed drawer template

oliverfoster:


moloko:
should it not be 
```
_globals._accessibility._ariaLabels
```
?

oliverfoster:
No, I've got to fix that later, the drawer, notify and some other mechanisms all work like this because these items don't have their own model: https://github.com/adaptlearning/adapt_framework/blob/develop/src/core/js/views/navigationView.js#L34

moloko:
+1 in that case

dancgray:
+1

taylortom:
+1

PR #72, opened 2015-03-19
Fix for x axis overflow under scrollbar

oliverfoster:
At the snap-points between mobile and tablet, tablet and desktop the scrollbar appears due to a slight overstretch of the content under the scrollbar. The fix effectively hides this overflow so that the bottom (x-axis) scrollbar never appears.

### DIFF
--- a/less/responsive.less
+++ b/less/responsive.less
@@ -1,6 +1,10 @@
 // Responsive layouts
 // ==================
 
+body {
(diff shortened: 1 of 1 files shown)


zenduo:
+1

taylortom:
+1

moloko:
+1

PR #75, opened 2015-04-20
Refactored, reordered colours, colour amends

guywillis:
Improvements to legibility of the CSS
Theme colour amends

### DIFF
--- a/less/variables.less
+++ b/less/variables.less
@@ -5,18 +5,18 @@
 // Main colors
 // ===========
 
-@primary-color:#3eb4e4;
-@secondary-color:#1b87b2;
-@tertiary-color:darken(#1b87b2, 10%);
+@primary-color: #00aeef;
+@secondary-color: darken(@primary-color, 20%);
+@tertiary-color: darken(@secondary-color, 20%);
 
-@foreground-color:#333;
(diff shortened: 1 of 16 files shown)


less/variables.less
```diff
@@ -309,66 +309,66 @@
...
 
-@notify-popup-padding:@notify-popup-padding-top @notify-popup-padding-right @notify-popup-padding-bottom @notify-popup-padding-left;
+@notify-popup-padding: @notify-popup-padding-top @notify-popup-padding-right @notify-popup-padding-bottom @notify-popup-padding-left;
 
 
 // Fonts
 // =====
 
-@import url(//fonts.googleapis.com/css?family=Open+Sans:300,400,600);
+@import url(https://fonts.googleapis.com/css?family=Roboto:300,400,700);
```
moloko: Please could you stick to a protocol-less URL like it was before? If you specify the protocol then you will get a secure/insecure content content issue - mainly in IE (which displays a confusing warning) but also I think Chrome too - if the content is run from server that uses a different protocol

oliverfoster:
+1

lc-thomasberger:
+1

dancgray:
+1

PR #80, opened 2015-06-11
Responsive layout

kirsty-hames:
- generic.less -- added device-widths and max-widths variables added.
- responsive.less removed -- code moved into adapt-defaults (was default-variables.less).
- page header added into template.
- content-max-widths added into templates to restrict content width (i.e. to allow a full width block but restrict content).
- page.less, article.less, block.less updated with new variables.


taylortom:
+1

PR #82, opened 2015-06-11
Narrative styles to replace component

dancgray:
Not sure if the padding or line hight should be separated out more into font and padding less files.


kirsty-hames:
+1

PR #83, opened 2015-06-12
new theme variables implemented

lc-thomasberger:
Accordion styles to replace component


taylortom:
+1

kirsty-hames:
+ 1 - The only note I have is do we need to set the font size/family, line height, color in here? Couldn't we just keep it in components.less rather than in each component's .less file. Seeing as it's just a basic theme just setting it in one place for all components would keep titles consistent.

PR #84, opened 2015-06-12
Components less

lc-thomasberger:
Component Title and Instruction Text updated to meet the AD:

- Removed .h4 from components template
- added Variables

### DIFF
--- a/less/src/component.less
+++ b/less/src/component.less
@@ -6,7 +6,13 @@
     }
 
     .component-title{
-        margin: @component-title-margin;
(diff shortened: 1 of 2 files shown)


taylortom:
+1

kirsty-hames:
+1 - I think we could do this for the rest of the theme too removing the 'h' classes styling. We put them in originality as a fallback styling whilst we were removing the 'h' tags for accessibility. Seeing as the accessibility is all in now and this is a major release I think we can drop them.

PR #85, opened 2015-06-12
buttons variables/LESS updated and transitions added

kirsty-hames:


less/src/buttons.less
```diff
@@ -3,29 +3,25 @@
...
     background-color: @button-color;
-    border: none;
     color: @button-text-color;
-    display: inline-block;
-    font-family: @font-family;
-    font-size: (@font-size*1.25);
-    font-weight: @font-weight-light;
-    line-height: (@font-size + @font-extra-line-height);
-    padding: @button-padding;
+    padding: @button-padding-top 10px @button-padding-bottom;
```
lc-thomasberger: do we need the 10px for side paddings?
i talked to Owen and the Text should be center alligned.

lc-thomasberger:
+1

PR #87, opened 2015-06-12
Text input less file

dancgray:
Not much to change in this one. Left some of the classes for input and labels in so widths can be changed.

### DIFF
--- a/less/src/textinput.less
+++ b/less/src/textinput.less
@@ -32,7 +32,7 @@
     }
     height: 28px;
     &.textinput-validation-error {
-        border: 2px solid @validation-error-border-color;
+        border: 2px solid @validation-error;
     }
 }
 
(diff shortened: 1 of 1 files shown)


less/src/textinput.less
```diff
@@ -0,0 +1,41 @@
...
+.textinput-item-textbox {
+    padding: 0 5px 0 5px;
+    width: 54%;
+    float: left;
+    .dir-rtl & {
+        float: right;
+    }
+    height: 28px;
+    &.textinput-validation-error {
+        border: 2px solid @validation-error-border-color;
```
kirsty-hames: - @validation-error-border-color needs to be updated to @validation-error.
- @item-border is an adapt-default variable. We could do with declaring this in the generic.less.

PR #90, opened 2015-06-12
Gmcq

lc-thomasberger:
Variables Implemented
Background-Color removed
option for 4-Column Layout (.col-4)

### DIFF
--- a/less/src/gmcq.less
+++ b/less/src/gmcq.less
@@ -17,7 +17,6 @@
 		        }
             }
         }
-
         &.odd {
 
             label {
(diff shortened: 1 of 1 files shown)


less/src/gmcq.less
```diff
@@ -0,0 +1,285 @@
...
+                }
+            }
+        }
+    }
+
+    &.col-4 {
+        .gmcq-item {
+            width: 25%;
+
+            &.item-0 {
```
kirsty-hames: isn't this the same as what's defined in lines 11-30?

PR #91, opened 2015-06-12
navigation LESS and variable update

kirsty-hames:


less/src/notify.less
```diff
@@ -32,24 +33,31 @@
...
+    .notify-popup-title-inner {
+        font-size: 36px;
+    }
+
     .notify-popup-button {
-        background-color: @notify-button-color;
-        color: @notify-button-text-color;
-        display: inline-block;
-        padding: @button-padding;
+        background-color: #fff;
```
lc-thomasberger: why hard coded colors?

dancgray:
+1

PR #92, opened 2015-06-12
Mcq less

dancgray:
Applies latest theme variables.

### DIFF
--- a/less/src/mcq.less
+++ b/less/src/mcq.less
@@ -1,0 +1,164 @@
+.mcq-component {
+    .mcq-item {
+        background-color:@item-color;
+        position:relative;
+    }
+
+    .mcq-item label {
+        color:@item-text-color;
+        display:block;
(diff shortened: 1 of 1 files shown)


less/src/mcq.less
```diff
@@ -0,0 +1,113 @@
+.mcq-component {
+    .mcq-item {
+        background-color:@item-color;
+        position:relative;
+    }
+
+    .mcq-item label {
+        color:@item-text-color;
+        display:block;
+        margin-bottom:2px;
```
lc-thomasberger: there is a variable to define distance between items @item-margin-top

kirsty-hames:
+1

@icon-correct-color and @icon-incorrect-color just need adding to the list of variables in color.less.

dancgray:
Changed the variables, but my test build is not showing validation-success, validation-error colour on the item icon. Will test some more.

PR #93, opened 2015-06-12
Matching

lc-thomasberger:


kirsty-hames:
As discussed I think repeated code which isn't likely to change can be removed. 

In regards to the browser specific styling fixes, should these be moved into the component itself?

PR #94, opened 2015-06-15
variables implemented

lc-thomasberger:
variables implemented
image width changed to 100%


kirsty-hames:
I think some of this code could be reduced. Do we really need to define code which is just a repeat of the default component styling? (i.e - .hotgraphic-graphic {position: relative} - this is just a repeat and isn't going to change).

PR #95, opened 2015-06-15
Slider less file

dancgray:


dancgray:
@kirsty-hames, sorry duplicate code removed

PR #96, opened 2015-06-15
navigation and drawer page level progress

kirsty-hames:
- basic styling added to override extension with updated variables.

### DIFF
--- a/less/src/page-level-progress.less
+++ b/less/src/page-level-progress.less
@@ -1,0 +1,97 @@
+// Navigation PLP
+// ==============
+
+.page-level-progress-navigation {
+    padding: 24px 36px 24px 18px!important; // Override defaults in extension
+    
+    .no-touch &:hover {
+
+        .page-level-progress-navigation-completion {
+            background-color: @navigation-progress-background-color-hover;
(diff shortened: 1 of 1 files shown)


less/src/page-level-progress.less
```diff
@@ -0,0 +1,93 @@
...
+    background-color: @navigation-progress-background-color;
+    border: 2px solid @navigation-progress-border-color;
+    border-radius: 50px;
+    .transition-background-color;
+}
+
+.page-level-progress-navigation-bar {
+    border-radius: 50px;
+    background-color: @navigation-progress-color;
+    transition: all 0.3s ease-in;
```
lc-thomasberger: do we need the transition all statement?

less/src/page-level-progress.less
```diff
@@ -0,0 +1,93 @@
+// Navigation PLP
+// ==============
+
+.page-level-progress-navigation {
+    padding: 24px 36px 24px 18px!important; // Override defaults in extension
```
lc-thomasberger: is it possible to use padding variables for that?

PR #97, opened 2015-06-15
Update transcript less

dancgray:
Do we need to add anything else? Would need some assets to change the play button.


lc-thomasberger:
+1

kirsty-hames:
+1 I think this should be fine as it is.

PR #107, opened 2015-06-30
Abu 650

taylortom:
The BIG theme update PR...


js/vanilla.js
```diff
@@ -1,5 +1,34 @@
...
+
+
+	// Page View
+	// =========
+
+	Adapt.on('pageView:postRender', function(view) {
+		var pageClasses = view.model.get('_classes');
+
+		if (pageClasses) {
+			this.$('.page').addClass(pageClasses);
```
lc-thomasberger: throws an error
this.$ is not a function

less/paddings.less
```diff
@@ -0,0 +1,109 @@
...
+@button-margin-bottom: @button-margin;
+
+// Drawer
+@drawer-item-padding-top: @item-padding;
+@drawer-item-padding-bottom: @item-padding;
+@drawer-item-padding-left: @item-padding;
+@drawer-item-padding-right: @item-padding;
+
+// Navigation
+// @navigation-padding-top: @navigation-padding;
```
lc-thomasberger: unused variables? remove when not used

less/src/adapt-defaults.less
```diff
@@ -1,3 +1,39 @@
+// IMPORTANT - THESE ARE THE DEFAULT VARIABLES / LESS FOR COMPATIBILITY WITH ADAPT OPEN SOURCE WHILST THEME IS IN WORKING PROCESS
+
+// THIS FILE WILL BE REMOVED BEFORE FINAL RELEASE - PLEASE COMMENT OUT VARIABLES AS THEY BECOME REDUNDANT
```
lc-thomasberger: i think we can't remove the file since some less variabels are only defined here and used in components less files. this will cause a less compile error.

eg @item-spacing is used in
  src\components\adapt-contrib-accordion\less\accordion.less
  src\components\adapt-contrib-hotgraphic\less\hotgraphic.less
  src\components\adapt-contrib-matching\less\matching.less

oliverfoster:
+1 with merge conflicts resolved (I'm going for the 'fix bugs later approach) have been using this as it's updated for a few weeks, it looks good.

taylortom:
+1

moloko:
+1

PR #115, opened 2015-07-14
ABU-905

lc-thomasberger:
removed padding-left from notify. not sure why it was added, but with it, the notify content is not center aligned to the screen.


taylortom:
+1

moloko:
+1

belal-almassri-spongeuk:
+1

zenduo:
+1

PR #119, opened 2015-07-14
ABU-956

lc-thomasberger:
font variables for body updated so that body text can be edited with variables from fonts.less


belal-almassri-spongeuk:
+1

zenduo:
+1

moloko:
+1

PR #128, opened 2015-07-15
ABU-896

lc-thomasberger:
removed variable @item-colour-inverted due to naming convention mismatch. replaced with @item-text-color.


zenduo:
+1

oliverfoster:
+1

kirsty-hames:
+1

PR #133, opened 2015-07-16
ABU-902

lc-thomasberger:
updated so that hotgraphic controlls and item count are vertically center aligned in the narrative popup toolbar


oliverfoster:
+1

zenduo:
+1

taylortom:
+1

PR #134, opened 2015-07-16
ABU-895

lc-thomasberger:
drawer updated so that disabled colors are not only applied on hover state.

when hovered over a disabled item, no hover effects are applied.


kirsty-hames:
+1

oliverfoster:
+1

zenduo:
+1

PR #136, opened 2015-07-16
ABU-779

lc-thomasberger:
forward and back arrows arrows are allways visible.
if this is not by design, then this PR will solve the Issue.

changed themes narrative.less so that narrative-hidden class is not overwritten.

### DIFF
--- a/less/src/narrative.less
+++ b/less/src/narrative.less
@@ -22,6 +22,10 @@
         text-decoration: none;
         background-color:@item-color;
 
+        &.narrative-hidden {
+            display: none;
+        }
+
         &:hover {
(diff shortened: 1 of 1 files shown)


kirsty-hames:
+1 maybe this should go in the component? This class is added within the component js so there must have been intention to hide these.

zenduo:
I agree with Kirsty, should go in the component.

lc-thomasberger:
good point. so do we need this PR still? (i think "narrative_focus_fix" will do the job)

kirsty-hames:
I think we can cancel this PR. Instead put the fix in for the narrative component and open a new PR.

PR #149, opened 2015-07-24
ABU-797

lc-thomasberger:
media queries added to article to reduce space between page hero image and content.

media queries and values to decrease padding the same as for blocks.


taylortom:
+1

kirsty-hames:
+1

moloko:
+1

PR #151, opened 2015-07-24
hide checkbox on ie8 and cleanup

lc-thomasberger:
removed duplicate styles.

ie8 does not apply opacity 0 to input elements, filter: alpha(opacity=0) added to fix this.


taylortom:
+1

kirsty-hames:
+1

moloko:
+1

PR #154, opened 2015-07-27
ABU-980

lc-thomasberger:
solves overlapping of text and icon by moving notify icon to right hand side for ltr and left hand side for rtl


taylortom:
+1

kirsty-hames:
+1

moloko:
+1

PR #155, opened 2015-07-28
ABU-1004

lc-thomasberger:
hides marking of question components in assessment

### DIFF
--- a/less/src/theme-extras.less
+++ b/less/src/theme-extras.less
@@ -92,3 +92,20 @@
 		background-image: none;
 	}
 }
+
+
+// Assessment - hide marking
+// requires a class called "assessment" on your assessment page
+// ==========
+
+ .assessment {
+	.buttons-marking-icon {
+		display: none;
+	}
+	.question-component {
+		.icon-tick,
+		.icon-cross {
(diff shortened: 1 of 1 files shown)


oliverfoster:
Did you not want to make this part of the assessment code? It seems sensible to have this feature in the extension.

lc-thomasberger:
the reason why I added it to the theme is because you have to add the assessment class by hand to the assessment page.

oliverfoster:
yea, i get that. 

i'm suggesting do this the standard way, having a switch in the authoring tool on the assessment article called _questions._canShowMarking: t/f ?  

that way people don't need to worry about class names or where to put them, just 'can the assessment questions show marking?' tick/untick.

lc-thomasberger:
Should work with all the question components since the class names for "buttons-marking-icon", "icon-tick" and "icon-cross" are used for every question component (at least tho ones used in the test package).
Will close this PR and add the styles to assessment.less.

taylortom:
+1

@oliverfoster I said to implement it this as we don't have time to do this properly prior to release - have added a JIRA feature request.

PR #156, opened 2015-07-29
ABU-989

lc-thomasberger:
this changes also affect ABU-979
removed duplicate styles
updated to use variable for margin definition
handling of width should be done in the component


taylortom:
+1

dancgray:
+1

kirsty-hames:
+1

PR #163, opened 2015-08-26
Abu 980

lc-thomasberger:
padding updated so that layout looks not so odd

question mark icon updated so that it is also left / right aligned for small screen widths

### DIFF
--- a/less/src/notify.less
+++ b/less/src/notify.less
@@ -81,11 +81,13 @@
     }
 
     .notify-popup-icon {
+        position: absolute!important;
         .icon {
             font-size: 60px;
             height: 60px;
             width: 60px;
             color: @notify-icon-color;
+            margin: 0!important;
         }
         right: 0px;
         left: inherit;
--- a/less/src/notify.less
+++ b/less/src/notify.less
@@ -26,11 +26,11 @@
         padding: @notify-padding-top 0 @notify-padding-bottom;
 
         @media all and (max-width: (@device-width-large + 20px)) {
-            padding: @notify-padding-top 20px @notify-padding-bottom;
+            padding: (@notify-padding-top / 2) 20px (@notify-padding-bottom / 2) !important;
         }
 
         @media all and (max-width: (@device-width-small + 20px)) {
-            padding: @notify-padding-top 10px @notify-padding-bottom;
+            padding: (@notify-padding-top / 3) 10px (@notify-padding-bottom / 3) !important;
         }
     }
 
@@ -89,6 +89,7 @@
         }
         right: 0px;
         left: inherit;
+        padding-bottom: (@notify-padding-bottom / 3);
         .dir-rtl & {
             right: inherit;
             left: 0px;


oliverfoster:
+1

moloko:
+1

guywillis:
What is the reasoning behind including 4 important tags into the less?

lc-thomasberger:
If i don't use the important tag I have to overwrite all the values in the theme. so we would end up with lot's of duplicate classes. I think in the theme it is ok to use !important.

less/src/notify.less
```diff
@@ -26,11 +26,11 @@
...
 
         @media all and (max-width: (@device-width-large + 20px)) {
-            padding: @notify-padding-top 20px @notify-padding-bottom;
+            padding: (@notify-padding-top / 2) 20px (@notify-padding-bottom / 2) !important;
         }
 
         @media all and (max-width: (@device-width-small + 20px)) {
-            padding: @notify-padding-top 10px @notify-padding-bottom;
+            padding: (@notify-padding-top / 3) 10px (@notify-padding-bottom / 3) !important;
         }
```
guywillis: The above two !important tags can be removed if the following snippet of CSS, located on line 15, is removed.

>.notify-type-prompt, 
    .notify-type-alert {
        .notify-popup-inner {
            padding: @notify-padding-top 0 @notify-padding-bottom;
        }
    }

guywillis:
What do you think about removing the position: static property from the .notify-popup-icon class in base.less rather than forcing a position: absolute on the same class within the theme? 

I don't recall it being used elsewhere, do you know if the static property is used elsewhere within notify?

Is the margin: 0 !important property required on the icon class? In my tests there is no difference between margin: 0 auto and the above.

lc-thomasberger:
Thanks for the Feedback Guy, I will have a look at your suggestions probably today.

I would prefer to leave the base.less untouched. The general approach is that we overwrite the default adapt styles in the theme.

How do you like the new Design of the Notify? So that the icon stays left or right aligned instead of center aligned on mobile devices. Another approach would be to hide the icon for small device widths.

guywillis:
I do like the new design of Notify, especially the right / left icon alignment. One thing to note with rtl support is that the question mark in Arabic is flipped as well. Do you think this is something we should include as part of this pull request?

I completely understand your preference, and agree with you, that we shouldn't be editing base.less freely. However, in this instance, the position: static has no effect on any other item within notify, as far as I'm aware, so we should be safe to remove it from base.less and then remove the overrides in this PR.

lc-thomasberger:
Good news, I could remove all the !important tags by nesting the selector containing the position absolute definition. So there is no need to update any styles in the base.less.

I assume that for rtl languages the icon will be replaced. If we have some sort of default question mark icon for rtl in the vanilla font I can update the theme so that this icon will be used for rtl courses.

guywillis:
Excellent, that sounds like a good solution. 

I don't believe we have an rtl version of the question mark icon in the font set. This issue is worth highlighting to the appropriate person to create this icon - do you know who that would be?

lc-thomasberger:
I am not really experienced with rtl languages.

PR #164, opened 2015-08-28
Abu-980

lc-thomasberger:
refactored to get rid of the important tags


guywillis:
Looks good Thomas.

I have a concern about the removal of the generic '.notify-popup-inner' class and listing it all under '.notify-type-prompt' and '.notify-type-alert'. 

I believe that this would break the Notify design on question feedback pop ups as I don't think these pop ups get either the prompt or alert class that popup-inner is now nested under. 

One solution would be to remove both the prompt and alert classes and list '.notify-popup-inner' on it's own, similar to how it was before. Do you agree?

PR #178, opened 2015-10-09
Drawer Font Size Amend

guywillis:
Set the font size of the drawer description to the same as the body.

Text was being overwritten by the <a> to <button> change.


lc-thomasberger:
+1

moloko:
+1

hrajotia:
+1

PR #188, opened 2016-03-03
Wrapped component.hbs with div tags

guywillis:
Surrounded component title, body, and instruction elements in a wrapper div + inner.

Added {{_component}} to wrapper divs to allow for specific component styling.

Un-themed divs by default, implemented to give course authors an easier time if they'd like to give component headings a background.

https://github.com/adaptlearning/adapt_framework/issues/965


moloko:
+1

oliverfoster:
+1

PR #193, opened 2016-06-28
modified to resolve issue #1073

chetan-hajare:


chetan-hajare:
The issue is actually https://github.com/adaptlearning/adapt_framework/issues/1073 not 1037

taylortom:
Is there any need for a disabled (but still visible) feedback button?

chetan-hajare:
Yes. when "_canShowFeedback" is set to true in assessment extension, it is needed to keep the feedback button disabled until submit button is clicked.
when "_canShowFeedback" is set to false, the submit button will occupy the full width and feedback button will not be visible.

moloko:
works for me in IE8, IE11, Firefox, Chrome and iOS 9

I wonder if `buttons-action-fullwidth` would be a better class name though?

brian-learningpool:
It feels a bit like too much of the `_canShowFeedback` logic here is in the Handlebars template.

Would this work better in the core as an updated to `buttonsView.js` in an `addClass()` call.

moloko:
@brian-learningpool seems OK to me in that there's no more complex logic going on here that in any other handlebars template... quite a bit less than some, in fact ;-)

moloko:
Although... now I think about it I see what you're saying. If we do this in the theme then every theme has to include this logic where really it should just be handling the styling. Adding the class via JavaScript probably _is_ the better way to go in this instance.

chetan-hajare:
The changes has to be in core framework not in theme so this pull request is invalid. I will make the changes in framework and raise the new pull request.

brian-learningpool:
Thanks @chetan-hajare.  I think your `buttons.less` changes were probably ok though?

chetan-hajare:
@brian-learningpool  - Do you mean to say css changes must be in theme's buttons.less and js code must be in frameworks's buttonView.js?

brian-learningpool:
Yes, the adding of the class should happen in buttonView.js, but the CSS changes were fine where there were.

chetan-hajare:
created a new PR - https://github.com/adaptlearning/adapt-contrib-vanilla/pull/194

PR #202, opened 2017-02-02
issue-1407 - changing less file name convention to match repo names

ciaranlp:
Changing component and extension less files naming convention in preparation to aid grunt task excluding files that are not required


ciaranlp:
I also combine the fonts.less from the less folder with fonts.less that was in the src folder

oliverfoster:
could you turn it into:
```
less/components/adapt-contrib-text/text.less
```
to allow for multiple files and set a prescendent

brian-learningpool:
So a structure like this?

- `/less`
  - `/components`
  - `/extensions`

With `components` and `extensions` treated as special folders and the .less files here only conditionally applied. 

@lc-thomasberger, do you have any thoughts?

lc-thomasberger:
I do like the idea of conditionally including a plugins less file. We need to make sure that custom themes that do not implement the folder structure will stay compatible with future versions of the Framework though.

The `/less` folder contains to many files in my opinion. The initial idea was to only have the files that contain variables in the first level. This should make customization easier for course creators without less skills.

Would suggest the following folder structure:´

- `/less`
  - `/components`
  - `/extensions`
  - `/base | shared | generic | ...`
  - `/colors.less`
  - `/padding.less`
  - `/fonts.less`
  - `/generic.less`

ciaranlp:
Hey @lc-thomasberger in the components folder would you keep the naming as it already is as in `accordion.less` or would you prefer `adapt-contrib-accordion.less`

oliverfoster:
please have a folder for each plugin, not a single filename. some plugins have multiple less files, like the media component - really anything with a third party library. i strongly disagree with having a single file per plugin if we're going to do exclusions like this.

oliverfoster:
how is this going to work with plugin dependencies?

lc-thomasberger:
@ciaranlp sub folders as suggested by @oliverfoster sounds very good.

brian-learningpool:
@ciaranlp , can I propose putting the plugin folders into a `/plugins` folder rather than separating out into components/extensions?

ciaranlp:
Can do if it helps the Grunt task

oliverfoster:
boxmenu needs moving into the plugins folder 👍

ciaranlp:
Should notify.less be in the plugins directory?

oliverfoster:
Not at the moment. It's still part of the core framework and not a plugin.

oliverfoster:
I would have called the ``defaults`` folder ``core`` and moved ``theme-extras`` out into a folder called ``custom`` or similar.

ciaranlp:
hey, @oliverfoster I don't mind doing this, although I don't think it makes much difference, so happy to go with the majority on this one!

danielstorey:
Think we're good to go! lol

moloko:
where are we at with this one? abandon or continue?

PR #203, opened 2017-03-08
Adds the ability to render attributes as part of the Notify

brian-learningpool:
Required to render 'data-adapt-id' as part of https://github.com/adaptlearning/adapt_framework/issues/1427.


templates/notify.hbs
```diff
@@ -1,5 +1,5 @@
 {{! Maintainers - Daryl Hedley}}
```
moloko: might as well take this out while you're here...

PR #204, opened 2017-03-16
select2 theme update 

lc-thomasberger:
Used styling from component less.
Removed styles that are not relevant for theming.


lc-thomasberger:
#1078

PR #205, opened 2017-03-29
issue-1493: removed redundant files

oliverfoster:
removed all redundant files which have been moved to core

fixes https://github.com/adaptlearning/adapt_framework/issues/1493


moloko:
just to point out that the templates in the framework currently aren't _exactly_ the same as the ones you're deleting... 

I've fixed that as part of framework PR [#1491](https://github.com/adaptlearning/adapt_framework/pull/1491)

You can see the differences [in this commit](https://github.com/adaptlearning/adapt_framework/pull/1491/commits/75acfed372510775882248b4bc3ebb48c9a4d85f)

Mostly the differences are minor, the only one that maybe isn't is [this one in navigation.hbs](https://github.com/adaptlearning/adapt_framework/pull/1491/commits/75acfed372510775882248b4bc3ebb48c9a4d85f#diff-df14f2f513e27161c6a051d4ed3d287dR6)

oliverfoster:
is this all good to proceed? @moloko  #1491 is merged

canstudios-louisem:
If these are being removed could less files like `mcq.less` be removed? as a lot of them are duplicates of the ones in their components?

oliverfoster:
Well... That's a long story. The short answer is probably no. They're there to make overriding easy.

Nothing in those files should be necessary for the functioning of the components, they should be style only. Ideally we'd have multiple plugins for the theme 'vanilla', including the core styling overrides, mcq override, slider etc as separate installable pieces. But we don't have plugin dependencies or packages sorted out quite yet. So at the moment t the core vanilla overrides are bundled with the vanilla theme plugin. If that makes any sense? It's not correct coding practice, but it does work for the moment.

canstudios-louisem:
Ok, that makes sense just wanted to check. It may also be worth thinking about the fact that now these are removed it's not obvious to someone editing the theme that template files can be edited at all let alone what they should contain.

Maybe update the docs at https://github.com/adaptlearning/adapt_authoring/wiki/Modifying-the-Vanilla-Theme and or https://github.com/adaptlearning/adapt_framework/wiki/Styling-your-course.

oliverfoster:
👍 wise plan

oliverfoster:
i've put in some placeholder text, perhaps someone could write a bit of clear and decent English in its place?

tomgreenfield:
Note: [wiki](https://github.com/adaptlearning/adapt-contrib-vanilla/wiki) needs an update too.

PR #206, opened 2017-03-31
updated vanilla fonts and less

ciaranlp:
Relates to  #issue-1498 updated vanilla font in Framework Core


moloko:
no longer required as fonts were removed in 3d5510b465ca65078661621d8bdb5ddcaca11fc5 - only need to be updated in core now

PR #207, opened 2017-04-27
Created body-text mixin

simondate:
A body text mixin can be used when overriding fields like attribution text if you wish to easily style it like regular text.

### DIFF
--- a/less/src/fonts.less
+++ b/less/src/fonts.less
@@ -100,6 +100,12 @@ body {
     line-height: @instruction-text-line-height;
 }
 
+.body-text {
+    font-size: @body-text-font-size;
+    font-weight: @body-text-font-weight;
+    line-height: @body-text-line-height;
+}
+
 .notify-popup-title {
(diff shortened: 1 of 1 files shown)


oliverfoster:
Can we have an issue explaining the requirement?

simondate:
I used it to style attribution text to look like regular body text.

oliverfoster:
you've defined a mixin but not used it anywhere or suggested it be documented anywhere.

simondate:
Do you want this mixin refrenced [here](https://github.com/adaptlearning/adapt-contrib-vanilla/wiki/Usage-and-Tips)?

I'm happy to have this PR closed if you think this is too niche.

oliverfoster:
I want to know how and where it is used. You need to be able to give others a reason to +1 by writing a use-case or giving an example. Mixins are used for all kinds of things, in this PR there is no explanation as to how this mixin can simplify or reduce styling overhead in Adapt.

PR #208, opened 2017-05-03
format link in menu body

lc-thomasberger:


less/src/boxmenu.less
```diff
@@ -37,6 +37,11 @@
 	.menu-body {
 		margin: @menu-block-body-margin-bottom 0;
 		color: @foreground-color-inverted;
+		
+		a {
+			text-decoration: underline;
```
tomgreenfield: The text decoration declaration is superfluous – this seems to be a hangover from `notify.less`.

PR #213, opened 2017-08-04
preview image for AT theme preview

canstudios-louisem:
Preview image for https://github.com/adaptlearning/adapt_authoring/pull/1692


canstudios-louisem:
We do I think they came from https://unsplash.com/. Would you prefer something like http://i.imgur.com/81jl8rs.jpg that is just a screenshot out of vanilla?

moloko:
Sorry for the slow response. If you can be sure the images came from unsplash and are definitely royalty-free/creative commons/similar then this is absolutely fine to use as-is

canstudios-louisem:
Yep they are the licence is [here](https://unsplash.com/license)

tomgreenfield:
Personally would prefer a screenshot as in Louise's comment above (looks like there's a custom component and non-standard colours in the mock-up).

moloko:
closing due to lack of follow up

PR #217, opened 2017-10-10
issue/1839: made boxmenu styling more specific

oliverfoster:
https://github.com/adaptlearning/adapt_framework/issues/1839

switched to using ``.boxmenu-menu`` instead of ``.menu``

### DIFF
--- a/less/src/boxmenu.less
+++ b/less/src/boxmenu.less
@@ -154,7 +154,7 @@
 
 @media all and (max-width: @device-width-medium) {
 
-	.boxmmenu-menu{
+	.boxmenu-menu{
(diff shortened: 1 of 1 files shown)


less/src/boxmenu.less
```diff
@@ -154,8 +154,8 @@
 
 @media all and (max-width: @device-width-medium) {
 
-   	.menu{
-   		.menu-inner {
+	.boxmmenu-menu{
```
moloko: typo - too many 'm's

oliverfoster: mmmm...

PR #218, opened 2018-04-11
Support for blockquote

simondate:
blockquote is a button that is used in CKEditor (the text editor used in the AT) so it makes sense to add support for it in vanilla.

Also fixed indentations and spacing before braces

### DIFF
--- a/less/src/component.less
+++ b/less/src/component.less
@@ -10,7 +10,7 @@
         }
     }
 
-    .component-title{
+    .component-title {
         margin-bottom: @component-title-margin-bottom;
         color: @title-color;
         text-align: left;
@@ -24,16 +24,22 @@
         padding: 0px;
     }
 
-    .component-body{
+    .component-body {
         margin: @component-body-margin-top 0px @component-body-margin-bottom;
 
         a {
-          text-decoration: underline;
-          color: @font-color;
+            text-decoration: underline;
+            color: @font-color;
+        }
+        
+        blockquote {
+            border-left: 2px solid @primary-color;
+            margin-left: 20px;
+            padding-left: 20px;
         }
     }
 
-    .component-body-inner{
+    .component-body-inner {
         padding: 0px;
     }
 
--- a/less/src/component.less
+++ b/less/src/component.less
@@ -31,12 +31,6 @@
             text-decoration: underline;
             color: @font-color;
         }
-        
-        blockquote {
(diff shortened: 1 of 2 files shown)


moloko:
for the benefit of those of us who can't render CSS in our heads, this is what it looks like when used in a text component:
blockquote

moloko:
looks good, my only thought is that this style only applies to component body text but you might also want to use it in the `.body` of a contentObject  - or (more likely) within question feedback.

simondate:
@moloko seems like a reasonable change. Do you know a good place to move it to?

I was thinking it could be added to either `page.less` or even into[base.less](https://github.com/adaptlearning/adapt_framework/blob/master/src/core/less/base.less) in core?

simondate:
Another option is making the quoted text _italicized_ . But I'm not sure if this is something that we should just leave to the course creator.

moloko:
> I was thinking it could be added to either page.less or even into base.less in core?

I'm **so** not the person to answer questions like that with any level of authority! Perhaps @guywillis or @lc-thomasberger could field that one?

> Another option is making the quoted text italicized . But I think this is something that we should just leave to the course creator.

I agree completely

lc-thomasberger:
I would definitely not put it into core, as quote styling is something that is highly custom and does not apply to all generic Adapt themes.
I would also not go for page.less as this holds the styling for the page itself. [theme-extras](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/master/less/src/theme-extras.less) is where I would put it. This styling should only act as an example that can be further customized.
You may of course als add a new less file with the name quote.less and add a variety of blog-quote-stylings here. 

The following does only apply to the existing vanilla theme, the structure will change.

simondate:
Ok, I'll move it to theme-extras for now.

moloko:
OK, looking good - the only problem now is that `@primary-color` isn't great in some situations:
blockquote_feedback
Is there an alternative that would work in a wider range of elements?

simondate:
@moloko would Notify be the outlier where it wouldn't work? We could just add an exception for it?

moloko:
@simondate also on boxmenu:
2018-05-04 18_08_20-adapt version 3 0 demonstration

It's not a big thing, happy to accept as is if it's a pain to sort... just thought I'd explore the possibility of another colour value that might work better in all situations but if there isn't one then fine

PR #219, opened 2018-05-03
Non-mobile replacing no-touch

simondate:
`no-touch` is problematic as it's applied on desktop operating systems that have a touch screen, which is a good proportion of Windows laptops released in the last three years. These users are more likely using a trackpad or mouse to navigate a course then their touchscreen - though this is open to debate.

This PR creates a variable which `os-android` and `os-ios` which makes >99% of the market share of mobile OS.

Original issue - https://github.com/adaptlearning/adapt_framework/issues/1820

### DIFF
--- a/less/generic.less
+++ b/less/generic.less
@@ -7,6 +7,10 @@
 @device-width-medium: unit(@adapt-device-medium, px);
 @device-width-large: unit(@adapt-device-large, px);
 
+// Mobile operating systems
+// ==========
+@non-mobile: ~"html:not(.os-ios):not(.os-android)";
+
 // Max-widths
 // ==========
 
--- a/less/generic.less
+++ b/less/generic.less
@@ -7,7 +7,7 @@
 @device-width-medium: unit(@adapt-device-medium, px);
 @device-width-large: unit(@adapt-device-large, px);
 
-// Mobile operating systems
+// Non-mobile operating systems
 // ==========
 @non-mobile: ~"html:not(.os-ios):not(.os-android)";
 
--- a/less/src/focus.less
+++ b/less/src/focus.less
@@ -1,7 +1,7 @@
 .focused,
 *:focus,
 input:focus + label {
-    .accessibility.no-touch & {
+    .accessibility@{non-mobile} & {
         outline: @focus-outline;
     }
 }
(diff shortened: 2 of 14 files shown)


simondate:
We've got a few MS surfaces in the office so I'll try to report back on how they work when they are free.

less/src/focus.less
```diff
@@ -1,7 +1,7 @@
 .focused,
 *:focus,
 input:focus + label {
-    .accessibility.no-touch & {
+    .accessibility@{non-mobile} & {
```
oliverfoster: `.accessibibilityhtml:not(.os-ios):not(.os-android)` isn't the intended selector

oliverfoster:
Can you make the sure the comment is indicative of the behaviour?
`Mobile operating systems` is more like `Not mobile operating systems`

less/src/focus.less
simondate: Should it be `html:not(.os-ios):not(.os-android).accessibibility` ?

oliverfoster: `@{non-mobile}.accessibility` should work fine

oliverfoster:
I'm happy that the code does what you want it to. I think I'm pretty happy with the approach. I'm not sure about the implications for the `.no-touch` class.
What does everyone else think?

moloko:
> `no-touch` is problematic as it's applied on desktop operating systems that have a touch screen, which is a good proportion of Windows laptops released in the last three years. 

Shouldn't this read "`touch` is problematic"?

less/generic.less
```diff
@@ -7,6 +7,10 @@
 @device-width-medium: unit(@adapt-device-medium, px);
 @device-width-large: unit(@adapt-device-large, px);
 
+// Non-mobile operating systems
+// ==========
+@non-mobile: ~"html:not(.os-ios):not(.os-android)";
```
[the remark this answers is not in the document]
simondate: We discussed here. https://github.com/adaptlearning/adapt_framework/issues/1820#issuecomment-385962492

Why not add other ones too?

moloko:
> Why not add other ones too?

Because Adapt only officially supports iOS (Safari), Android (Chrome) and Windows Phone (IE11 Mobile)...

oliverfoster:
The more I look at the code, the more I think it should definitely be in the core and not in the theme.

oliverfoster:
closed as very out of date

PR #220, opened 2018-06-18
accessibility: issue/#2137 reverted to browser default focus rectangle

oliverfoster:
[#2137](https://github.com/adaptlearning/adapt_framework/issues/2137)
* Reverted to browser defaults

### DIFF
--- a/less/generic.less
+++ b/less/generic.less
@@ -16,8 +16,6 @@
 // Borders and Outlines
 // ====================
 
-@focus-outline: 0; //focus outline is displayed when accessibility is enabled
-
 @button-border: 0px solid @button-border-color;
 
 @item-border: 3px solid @item-border-color;
--- a/less/generic.less
+++ b/less/generic.less
@@ -16,7 +16,7 @@
 // Borders and Outlines
 // ====================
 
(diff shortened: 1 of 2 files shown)


less/generic.less
```diff
@@ -16,7 +16,7 @@
 // Borders and Outlines
 // ====================
 
-@focus-outline: 3px solid orange; //focus outline is displayed when accessibility is enabled
+@focus-outline: 1px; //focus outline is displayed when accessibility is enabled
```
tomgreenfield: As per [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/outline):
> Note: The outline will be invisible if its style is not defined. This is because the style defaults to none.

Suggest either choosing an outline style for now or set to `0` if the intention is to hide it.

Going forward, we'll need to remove all `outline` CSS overrides before the browser default for `:focus` is used.

tomgreenfield:
Suggest this variable is just removed completely? It's already defined in [core](https://github.com/adaptlearning/adapt_framework/blob/accessibilityv4/src/core/less/base.less#L2).

PR #222, opened 2018-07-05
accessibility: issue/#2083 updated plp styling for completion indicators

oliverfoster:
[#2083](https://github.com/adaptlearning/adapt_framework/issues/2083)
* Made progress bars uniform in colour and size for dark and light backgrounds
* Added styling for new plp progress bars
* Added inner border to make bars easier to read
* Unified styling under new class name

### Test with
[plp#107](https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/pull/107)

List of outstanding PRs: https://github.com/adaptlearning/adapt_framework/issues/2206.

### DIFF
--- a/less/colors.less
+++ b/less/colors.less
@@ -127,9 +127,9 @@
 // Menu
 @menu-item-color: @primary-color;
 @menu-item-text-color: @primary-color-inverted;
-@menu-item-progress-color: darken(@menu-item-color, 20%);
-@menu-item-progress-fill-color: @menu-item-text-color;
-@menu-item-progress-border-color: transparent;
+@menu-item-progress-color: @navigation-color;
+@menu-item-progress-fill-color: @navigation-icon-color;
+@menu-item-progress-border-color: @navigation-icon-color;
 
 // Notify
 @notify-color: @primary-color;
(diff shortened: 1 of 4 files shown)


oliverfoster:
I'm not sure the styling difference between content and PLP drawer is an issue. Harmonizing them isn't going to work as the drawer is so dark.

The only thing I can think of doing is allowing the background to show through, so the fill is uniform rather than the colour.
image
image

oliverfoster:
Have settled with this, as requested by @tomgreenfield and @moloko and @lc-thomasberger together:
image
image
image
image

nilslp:
The progress bars look slightly like toggles when there is no completion on them, I would think it would be better to not have any fill if it has no progress. But the outline etc is a great job 👍 

https://user-images.githubusercontent.com/7974663/42630199-d6f02c42-85cd-11e8-9600-e0d1c6061249.png

moloko:
@nilslp I think in the drawer they're wide enough they don't look like switches but agree that on the component title they could be confused with switches. I think it would only be a brief moment of confusion though - in that it would be pretty obvious that they're not clickable and as soon as you saw one fill in you'd realise what they were. Perhaps having a tooltip along the lines of 'you have/have not completed this component' might help (for desktop browser users anyway)?

oliverfoster:
The latest styling changes these to have a flat end:
image

Definitely less toggley.

oliverfoster:
need 2x +1 to get this merged when needed.

PR #231, opened 2019-02-22
Expose all colour variables to authoring tool editor

canstudios-nicolaw:
Issue: https://github.com/adaptlearning/adapt_framework/issues/2359


moloko:
sorry bit of an afterthought - this deletes the `_screenSize` object, just wanted to check that's correct?

PR #235, opened 2019-08-16
v5: add background-color to drawer__toolbar

chucklorenz:
This is related to framework issue #2502.

Currently the drawer-toolbar has a transparent background. This PR assigns the @drawer-color to it background. This solid background i necessary to allow items to scroll behind it unseen.


guywillis:
I don't believe this PR is required anymore due to drawer padding top being moved into the `.drawer__inner`.

PR #236, opened 2019-09-02
#2512

simondate:


simondate:
See https://github.com/adaptlearning/adapt_framework/issues/2512

oliverfoster:
https://github.com/adaptlearning/adapt-contrib-vanilla/blob/release/v5/less/_defaults/_font-config.less#L10

It might be better to use `@title-family` ?

PR #241, opened 2020-01-09
img zoom enable class added

guywillis:
https://github.com/adaptlearning/adapt_framework/issues/2615

### DIFF
--- a/less/plugins/adapt-contrib-accordion/accordion.less
+++ b/less/plugins/adapt-contrib-accordion/accordion.less
@@ -54,7 +54,7 @@
     border-radius: 5px;
   }
 
-  .accordion.has-img-zoom &__item-image-container {
+  &.has-img-zoom &__item-image-container {
     .img-zoom();
   }
 }
--- a/less/plugins/adapt-contrib-accordion/accordion.less
+++ b/less/plugins/adapt-contrib-accordion/accordion.less
@@ -54,7 +54,7 @@
     border-radius: 5px;
   }
 
-  &__item-image-container {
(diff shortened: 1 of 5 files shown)


guywillis:
Hi @cahirodoherty-learningpool,

It is intentional as it allows us to target the `.has-img-zoom` class specifically at the component level. If the `.has-img-zoom` class didn't have the associated component class then there is a chance all the components on a page would have their img zooms enabled if a user were to add a zoom effect to the box menu item via the contentObject class field.

```
.accordion {
  .accordion.has-img-zoom &__item-image-container {
    .img-zoom();
  }
}
```

compiles as 

```
.accordion.has-img-zoom .accordion__item-image-container {
 // code here
}
```

cahirodoherty-learningpool:
Hi Guys, thanks for your reply. I understand now the CSS that gets compiled.
One thing I would like to suggest is perhaps replacing the inner `.accordion` with an ampersand? To me, at least, this would prove more readable and consistent with the less as-is.

e.g.:
```
.accordion {
  &.has-img-zoom &__item-image-container {
    .img-zoom();
  }
}
```

guywillis:
Agreed, I'll make the changes across the PRs

guywillis:
@cahirodoherty-learningpool Amends have been made. The only exception is hotgraphic due to the `.hotgraphic` class being applied at the notify level

cahirodoherty-learningpool:
> Agreed, I'll make the changes across the PRs

Thanks Guy. A little pedantic, I know, but I think it looks better now 👍

PR #253, opened 2020-03-03
Removed schema theme picker defaults

guywillis:
Resolves https://github.com/adaptlearning/adapt_framework/issues/2609

### DIFF
--- a/less/_defaults/_colors.less
+++ b/less/_defaults/_colors.less
@@ -14,6 +14,11 @@
 @transparent-dark: transparent;
 // --------------------------------------------------
 
+// Note: If customising a theme for the Authoring Tool please update the
+// corresponding default value for that variable in the properties.schema file.
+// This is to stop any amended variables below being overwritten by
+// the Theme Picker in the Authoring Tool.
+
 // --------------------------------------------------
 // Body, link, and heading colour
 // --------------------------------------------------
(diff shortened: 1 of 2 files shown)


canstudios-nicolaw:
Is the duplication in the export a big problem? 

"An unfortunate consequence of removing these defaults is that the Theme Picker interface in the AAT no longer pre-populates itself with the standard vanilla theme colours." - This seems like quite a noticeable and negative difference to the AT UI to me, I think it will make things look quite broken.

guywillis:
The duplication can become a problem if you are customising a themes colour variables (such as `@item-color`) and uploading a new theme. Without the additional process of either removing the duplicates from the schema or updating them to reflect the new colour hex codes users can find themselves reverting colours back to the defaults set by Vanilla by accident (by pressing 'Restore to preset' button in the Theme Picker area of the AAT).

canstudios-nicolaw:
Personally I think that theme authors should be setting the default values when they build a theme if they want it to work well in the AT. This should probably be part of theme-building docs.

This change will have a negative impact on AT-only users that want to tweak the colours in vanilla.

Is there a potential for a middle-ground somewhere? Could the theme picker UI do anything differently that would make it harder to revert the colours back to the wrong defaults if the theme author hasn't set them?

guywillis:
I agree that this change is far from ideal and could have a negative impact on AT users.

Ideally, the Theme Picker could find the hex codes for the predefined variables so AT users wanting to create custom themes wouldn't have to set defaults - the variables in theme itself (_colors.less) would be the defaults. Do you think there is potential to do this in the AT?

PR #263, opened 2020-07-02
resolves #2821

kirsty-hames:
resolves https://github.com/adaptlearning/adapt_framework/issues/2821

### DIFF
--- a/example.json
+++ b/example.json
@@ -97,10 +97,13 @@ Content Object classes
 
 "_classes": "header-color",
 "_comment": "Sets up the header solid background colour mixin. Combine with one of the following classes e.g. 'header-color background'. Note: the following class must be a predefined variable. Mixin can be extended to include client colours. Can be set for menu and/or page header.",
+
 "_classes": "transparent-light",
 "_comment": "Sets the background colour of the header to transparent and changes the font colour to @font-color. Useful for displaying text over a light background image.",
+
 "_classes": "transparent-dark",
 "_comment": "Sets the background colour of the header to transparent and changes the font colour to @font-color-inverted. Useful for displaying text over a dark background image.",
+
 "_classes": "background",
 "_comment": "Sets the background colour of the header to use the variable @background (default is black) and changes the font colour to use the variable @background-inverted (default is white). These variables can be changed in the ‘_colors.less’ file.",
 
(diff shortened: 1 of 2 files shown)


less/project/theme-extras.less
```diff
@@ -18,6 +18,38 @@
...
+// e.g. 'header-color transparent-dark'
+// Note: both arguments must be predefined variables
+// --------------------------------------------------
+.header-color-mixin(background, background-inverted);
+.header-color-mixin(transparent-light, font-color);
+.header-color-mixin(transparent-dark, font-color-inverted);
+
+.header-color-mixin(@color, @color-inverted) {
+
+.header-color.@{color} {
```
cahirodoherty-learningpool, line 32: Minor: Would it be worth moving the `@` reference out of the mixin and onto where it is called instead?
I think this would read clearer when passing `@font-color` etc.
e.g.
`.header-color-mixin(@transparent-light, @font-color);`

kirsty-hames: Hey @cahirodoherty-learningpool, thanks for your feedback. I kept the `@` reference in the mixin for consistency with the existing `block-color` mixin. I agree that your approach would read clearer but I think it was implemented this way so that the variable name would double up as the colour class. Otherwise we would need to define a class for each colour. @guywillis was this your initial thinking?

cahirodoherty-learningpool: Okay. In that case I am happy for the consistency to be maintained 👍

guywillis: @cahirodoherty-learningpool @kirsty-hames 

That's correct - the double `@` symbol within the mixin was implemented so that a singular name could be used for both a class and a variable. Agreed that it's not the cleanest thing to read in the Less, however, the mixin will make any value passed into it a variable due to the double `@@` symbols.

example.json
```diff
@@ -95,6 +95,15 @@ Content Object classes
 "_classes": "hide-page-header",
 "_comment": "Hides the page header on the page",
 
+"_classes": "header-color",
+"_comment": "Sets up the header solid background colour mixin. Combine with one of the following classes e.g. 'header-color background'. Note: the following class must be a predefined variable. Mixin can be extended to include client colours. Can be set for menu and/or page header.",
+"_classes": "transparent-light",
```
guywillis, line 100: Very minor point - could a new line be added between each entry to make it a little easier to read please

PR #270, opened 2020-08-20
Modify slider.less styles

joe-replin:
I have one very low-contrast monitor at home and with it, I'm having a difficult time seeing the slider fill and handle. The handle is the active button, so I believe it should take on the primary color of the theme. This causes a little conflict with the range numbers being the same color, so I've changed the selected range color to instead include a rounded border with the primary-color. Lastly, I've changed the fill color to the greyDark color in the theme for better contrast.

### DIFF
--- a/less/plugins/adapt-contrib-slider/slider.less
+++ b/less/plugins/adapt-contrib-slider/slider.less
@@ -21,9 +21,12 @@
   }
 
   &__number-selection {
-    background-color: @item-color;
-    color: @item-color-inverted;
+    background-color: @item-color-inverted;
+    border: 0.125rem solid @item-color;
+    box-sizing: border-box;
+    color: @item-color;
     z-index: 5;
+    line-height: 2.25rem;
   }
 
   &__number-model-answer {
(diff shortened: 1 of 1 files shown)


less/plugins/adapt-contrib-slider/slider.less
```diff
@@ -21,9 +21,12 @@
   }
 
   &__number-selection {
```
guywillis, line 23: Suggest removing the additional styles and instead change the variables colours alongside the introduction of a box-shadow style that does not impact the elements height.

```
&__number-selection {
  background-color: @font-color-inverted;
  color: @font-color;
  z-index: 5;
}

&:not(.is-complete) &__number-selection {
  box-shadow: 0 0 0 2px @item-color inset;
}
```

less/plugins/adapt-contrib-slider/slider.less
```diff
@@ -67,11 +70,12 @@
   }
 
   &__fill {
-    background-color: @item-color;
+    background-color: @greyDark;
```
guywillis, line 73: A hard coded variable here would cause the slider fill to never be editable in the Authoring Tool. How does the slider fill look when both the handle and fill are set to use `@item-color`?

guywillis (changes requested):

oliverfoster:
stale

PR #295, opened 2021-07-26
issue/3177 column structure

guywillis:
Looks to resolve https://github.com/adaptlearning/adapt_framework/issues/3177

* initial implementation of column structure as custom css classes

### DIFF
--- a/less/_defaults/_spacing-containers.less
+++ b/less/_defaults/_spacing-containers.less
@@ -36,12 +36,11 @@
 // Block padding
 // --------------------------------------------------
 @block-padding-ver: 4rem;
-@block-padding-hoz: 1rem;
+@block-padding-hoz: .5rem;
 
 // Component padding
 // --------------------------------------------------
-@component-padding-ver: 1rem;
-@component-padding-hoz: 1rem;
+@component-padding-hoz: .5rem;
 
 // Notify padding
 // --------------------------------------------------
--- a/less/_defaults/_spacing-containers.less
+++ b/less/_defaults/_spacing-containers.less
@@ -36,11 +36,12 @@
 // Block padding
 // --------------------------------------------------
 @block-padding-ver: 4rem;
-@block-padding-hoz: .5rem;
+@block-padding-hoz: 1rem;
 
 // Component padding
 // --------------------------------------------------
-@component-padding-hoz: .5rem;
+@component-padding-ver: 1rem;
+@component-padding-hoz: 1rem;
 
 // Notify padding
 // --------------------------------------------------
--- a/less/_defaults/_spacing-containers.less
+++ b/less/_defaults/_spacing-containers.less
@@ -18,6 +18,8 @@
 @menu-header-padding-ver: 2rem;
 @menu-header-padding-hoz: .5rem;
 
+@menu-item-container-margin: 1.5rem;
+
 // Page container and header padding
 // --------------------------------------------------
 @page-padding-ver: 0;
--- a/less/core/component.less
+++ b/less/core/component.less
@@ -20,12 +20,12 @@
   }
 }
 
-.component.is-left + .component.is-right {
-  // Converted media query to use min width to sync responsive trigger with js
-  // https://github.com/adaptlearning/adapt_framework/issues/2865
-  padding-top: @block-padding-ver;
+// .component.is-left + .component.is-right {
+//   // Converted media query to use min width to sync responsive trigger with js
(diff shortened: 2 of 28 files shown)


oliverfoster:
could you remove unnecessary comments?

less/core/component.less
```diff
@@ -20,12 +20,12 @@
...
+//   padding-top: @block-padding-ver;
 
-  @media (min-width: @device-width-medium) {
-    padding-top: 0;
-  }
-}
+//   @media (min-width: @device-width-medium) {
+//     padding-top: 0;
+//   }
+// }
```
oliverfoster, line 31: I this section needed?

less/_defaults/_spacing-containers.less
```diff
@@ -36,11 +36,12 @@
 // Block padding
 // --------------------------------------------------
 @block-padding-ver: 4rem;
-@block-padding-hoz: .5rem;
+@block-padding-hoz: 1rem;
 
 // Component padding
 // --------------------------------------------------
-@component-padding-hoz: .5rem;
+@component-padding-ver: 1rem;
```
kirsty-hames, line 43: @guywillis, any particular reason for increasing the padding to 1rem? I'd be inclined to keep this as .5rem as it aligns the content with the header for consistency. Also I think we purposely reduced the content padding in v5 to allow more room for content (1rem in total between left/right components on desktop, .5rem horizontal padding on mobile).

less/project/columns.less
```diff
@@ -0,0 +1,32 @@
+.make-columns(@size, @media-query, @count, @iteration: 1) when (@count >= @iteration) {
```
kirsty-hames, line 1: In my initial prototype for this I included a 0 column width for instances where we might want to include a blank component to create white space for specific view widths. I can't see the need for a 0 width other than with a blank component so we could target 'blank' exclusively for this. I'll have a look into this and make a suggestion...

less/project/theme-extras.less
```diff
@@ -157,41 +157,87 @@
...
-  .narrative {
-    &__progress {
-      background-color: fade(@@color-inverted, 50%);
+}
+
+// --------------------------------------------------
+// Component
+// --------------------------------------------------
+
+// Component header bg color mixin
```
kirsty-hames, line 202: The addition of component header style and various style/spacing tweaks could be moved out into a separate PR as it isn't column related?

kirsty-hames (commented):
The move to 12 column layout support is a big shift from the two column layout we're all use to with Adapt so I think a visual example and some documentation around this would be good to share. I've got my initial research, ramblings and a demo course I can tidy up.

less/_defaults/_spacing-containers.less
guywillis: @kirsty-hames Increased the padding as, when combined with either `bg-color` or `comp-header-bg-color`, 1rem looked nicer.

I think we could update the other existing .5rem to 1rem so they're all in alignment - observationally, 8px gap between the edge of the content and the edge of the screen is pretty small and increasing this width, in my view, makes the small screen display more appealing. 

Ultimately, I think we should look to increase the content display area for larger screens which should negate the increase in padding.

less/project/columns.less
guywillis: There's some usefulness to it but it's definitely a small window with the current device width break points. When we increase these break points the window for this use case will increase.

By default all components are 100% width. With the xs 0 column width selector you could change the size of a component between 0 and the sm device width small selector (or larger) to be a value that's not 100%.

less/project/theme-extras.less
guywillis: Agreed with splitting into a separate PR.

It was included for the time being so 1) it didn't get lost / forgotten about and 2) helpful for development

less/_defaults/_spacing-containers.less
kirsty-hames: I agree that `comp-header-bg-color` looks better with the increase of padding but seeing as this is an optional style I think the padding should be specific to this style and not global. 
I know it's personal preference but I've never had a request to increase the horizontal padding on projects. For mobile in particular I find more space for content is preferred.

guywillis: I would agree but I fear the potential mismatch with some components with .5 padding and some set at 1rem depending on whether they have a custom class applied or not - the idea behind the change is to ensure consistency at the expense of a few pixel padding.

oliverfoster:
bump @guywillis

oliverfoster (approved):

PR #302, opened 2021-09-23
add option to background-image size

### LINKED ISSUE #301: add '100% 100%' as background-image size option
Proposing that we add '100% 100%' to the AT schema as an option for background-image size.

Current options in schema are 'auto', 'cover', and 'contain'. CSS also allows percentages. '100% 100%' stretches the image to fill the space. It will distort images whose dimensions are not in the same ratio; but is helpful for graphic content that can bear it. Works especially well with SVGs. Without this option in the AT, background-image size must be specified in the uploaded theme or in Custom CSS/LESS code.

chucklorenz:
closes #301

Adds "100% 100%" to schemas, amends help in schemas, amends README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,8 +52,8 @@ The **Vanilla** theme also exposes [*colour variables*](https://github.com/adapt
 >>**\_backgroundRepeat** (string): This attribute defines how the background image repeats. Properties include **repeat**, **repeat-x**, **repeat-y** and **no-repeat**.
 Repeat-x: The background image is repeated only horizontally. Repeat-y: The background image is repeated only vertically.
 
->>**\_backgroundSize** (string): This attribute defines the size the background image display. Properties include **auto**, **cover** and **contain**.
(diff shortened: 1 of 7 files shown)


cahirodoherty-learningpool (approved):
👀

guywillis (approved):
👀  Looks great, thanks Chuck.

guywillis:
Would you mind also creating a PR for the boxmenu to do the same thing? https://github.com/adaptlearning/adapt-contrib-boxMenu/blob/master/properties.schema#L79

I think the version could be bumped too 5.7.2 rather than 5.8.0 as it's a small addition rather than new feature.

chucklorenz:
Yes, @guywillis, I can do the same for boxMenu.

kirsty-hames (changes requested):
Looks good thanks but I agree with bumping version to 5.7.2 rather than 5.8.0.

kirsty-hames (approved):

Issue #311, opened 2021-11-12
Matching inconsistent BEM naming


guywillis:
Related to matching issue [139](https://github.com/adaptlearning/adapt-contrib-matching/issues/139)

PR #312, opened 2021-11-12
Matching BEM naming corrections

guywillis:
Resolves https://github.com/adaptlearning/adapt-contrib-vanilla/issues/311

* Converted matching naming so each item is treated as a block within BEM


cahirodoherty-learningpool (approved):
👀

eleanor-heath (approved):

oliverfoster (approved):

Issue #317, opened 2021-11-26
Inline icon support


guywillis:
Allow support for icons displays inline alongside body text. 

Any icon defined within the vanilla icon set can be used - see here https://github.com/adaptlearning/adapt-contrib-vanilla/wiki/Icons

Note: be sure to add spaces before and after the div to ensure consistent spacing between text and icon.

e.g.

JSON
`"body": "Lorem ipsum dolor sit amet, consectetur <div class='inline-icon icon icon-controls-down'></div> adipiscing elit. Curabitur nec urna et nisl congue porta.",`

Visual
<img width="822" alt="Screenshot 2021-11-26 at 14 40 16" src="https://user-images.githubusercontent.com/11569678/143597081-728873b7-8f40-4b6b-8df8-ce1c406996c0.png">

PR #318, opened 2021-11-26
Inline icon support

guywillis:
Resolves https://github.com/adaptlearning/adapt-contrib-vanilla/issues/317

* Add support for icons to appear inline with text - can be title / body / instruction


ethan-lp (approved):

cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

PR #320, opened 2022-01-13
issue/319 Fixed es6 conversion bug

### LINKED ISSUE #319: Page header background images not working with Vanilla 6
Having recently updated to Framework 5.18.1, which pulled in the upgrade to Vanilla 6.0.0 (from 5.7.0), we discovered that page header background images were not being included in the course.

Breakpointing in the debugger suggests that Theme#initialize() is not being called, which prevents onPostRender() hook (which creates the ThemePageView etc. objects) from firing.

Changing the class declaration from
```
export default class Theme extends Backbone.Controller {
  ...
}
```
to

```
class Theme extends Backbone.Controller {
  ...
}

export default new Theme();
```
(in alignment with other Adapt plugins updated to use ES6-classes) seems to make things go again.

oliverfoster:
fixes #319 

### Fixed
* Main singleton instance created on load

### DIFF
--- a/js/theme.js
+++ b/js/theme.js
@@ -4,7 +4,7 @@ import ThemeArticleView from './themeArticleView';
 import ThemeBlockView from './themeBlockView';
 import ThemeView from './themeView';
 
-export default class Theme extends Backbone.Controller {
+class Theme extends Backbone.Controller {
 
   initialize() {
     this.listenTo(Adapt, {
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

taylortom (approved):

eleanor-heath (approved):

PR #330, opened 2022-05-18
move component padding styles

### LINKED ISSUE #326: Is standardized padding in the most logical file?
While working with Vanilla v6.1.0, I had to track down component padding. I encountered the new standardized padding (#321). I have no issue with the change. But I was confused by where I found the code. Note that I was searching for the padding while in my local environment. I found it in _columns.less_. That made me scratch my head. I wasn't using columns. And it made me wonder if there was a collision/side-effect coming from the implementation of columns. I was about to report the issue when I discovered the existing issues regarding standardized padding and educated myself.

If I understand the rationale behind the standardized padding, it seems that a more logical place for this Less code would be in Vanilla/less/**core**/component.less. Placing it in [Vanilla/less/**project**/columns.less](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/master/less/project/columns.less) gives the impression that the padding styles would be invoked only when columns are used.

chucklorenz:
Resolves #326 
Moves padding styles from project/columns.less to core/component.less.

### DIFF
--- a/less/core/component.less
+++ b/less/core/component.less
@@ -1,4 +1,14 @@
 .component {
+  &__container {
+    flex-direction: unset;
+    margin-left: -@component-padding-hoz;
+    margin-right: -@component-padding-hoz;
+  }
+
+  &__inner {
+    padding: @component-padding-ver @component-padding-hoz;
+  }
+
   &__title {
     margin-bottom: @component-title-margin;
     .component-title;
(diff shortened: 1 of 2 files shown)


cahirodoherty-learningpool (approved):
👀

guywillis (approved):
👀

oliverfoster (approved):

PR #354, opened 2022-11-03
Update: component vertical alignment property (fixes #353)

### LINKED ISSUE #353: Consolidate component alignment custom classes into property
At present, a user is required to type out the custom class required for changing the component vertical alignment within a block. This custom class has to be applied to the _classes field on the block or corresponding field within the authoring tool.

The application of the vertical alignment class could be optimised by converting the options to a drop down selection, similar to the vanilla's background image styles property. 

### Your environment
* AAT / FW

guywillis:
Fixes: https://github.com/adaptlearning/adapt-contrib-vanilla/issues/353

### Update
* Consolidates the vertical alignment of components within a block into a json property that applies alignment classes on the block depending on property value

### DIFF
--- a/example.json
+++ b/example.json
@@ -83,8 +83,18 @@ blocks.json
         "_small": "class-small"
     },
     // Divider Block - adds class to `.block` element
-    "_isDividerBlock": false
+    "_isDividerBlock": false,
+    // Defines the vertical alignment of the child component(s) in relation to the block.
+    // Top: Aligns descendents to the top of the block. Center: Aligns descendents to the center of the block. Bottom: Aligns descendents to the bottom of the block.
+    // The default alignment is `top`.
+    "_componentVerticalAlignment": "center"
 }
+
+"_classes": "align-vert-center",
+"_comment": "Aligns the blocks child components centrally on the vertical axis.",
+
+"_classes": "align-vert-bottom",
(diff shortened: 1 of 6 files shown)


oliverfoster (approved):

kirsty-hames (changes requested):
Config looks good thanks @guywillis but I'm not sure the class styles need moving from _theme-extras.less_ to _block.less_. I think the current setup of having a file containing all the extra optional styles is an intuitive place to look and making the class a config setting doesn't change that this is still an optional style.
Although this PR only relates to the vertical alignment property, if more classes are to become config options to improve usability in the AAT then it could get messy scattering classes across core/plugin less files? Happy to discuss or be told otherwise.

joe-allen-89 (approved):

kirsty-hames (approved):

kirsty-hames:
> Config looks good thanks @guywillis but I'm not sure the class styles need moving from _theme-extras.less_ to _block.less_. I think the current setup of having a file containing all the extra optional styles is an intuitive place to look and making the class a config setting doesn't change that this is still an optional style. Although this PR only relates to the vertical alignment property, if more classes are to become config options to improve usability in the AAT then it could get messy scattering classes across core/plugin less files? Happy to discuss or be told otherwise.

After discussion, seeing as this class is exclusive to block model only, it makes sense this is contained to block.less.

PR #355, opened 2022-11-07
Fix: Make Notify icon buttons dark in initial state (fixes #350)

### LINKED ISSUE #350: Notify popups - disabled pagination button styles
### Subject of the issue/enhancement/features
Pending the approval of https://github.com/adaptlearning/adapt-contrib-hotgraphic/pull/242, disabled pagination buttons in Hotgraphic popups will now be visible.  This leads to the question of what they should look like.  Currently, they will inherit disabled button styles which may or may not look out of place.

Thoughts on the below screenshots?  I prefer the lowered opacity look, but it may not be ideal for low vision users.

**Current view with disabled buttons hidden:**
hotgraphic-disabled-button-hidden

**After removing visibility: hidden**
hotgraphic-disabled-button-visible

**Alternate style with lowered opacity**
hotgraphic-disabled-button-translucent

### Your environment
* Framework 5.x

swashbuck:
Fixes #350 

### Fixes
* Makes Notify icon buttons darker by default. Using the same values as previous hover state.
* New hover state would be slightly darker. From my understanding, hover states do not need to pass accessibility contrast requirements (citation needed...).
* Affects "paging" buttons in a popup as well as the close button.
* Disabled buttons - no change. These use default disabled button styles.

### Testing
1. Click on a pin in a Hot Graphic component.
2. Interact with the buttons.

### DIFF
--- a/less/_defaults/_colors.less
+++ b/less/_defaults/_colors.less
@@ -132,10 +132,10 @@
 @notify-btn-hover: darken(@notify-btn, 10%);
 @notify-btn-inverted-hover: @notify-btn-inverted;
 
-@notify-icon: @notify;
+@notify-icon: darken(@notify, 10%);
 @notify-icon-inverted: @notify-inverted;
 
-@notify-icon-hover: darken(@notify-icon, 10%);
+@notify-icon-hover: darken(@notify, 15%);
 @notify-icon-inverted-hover: @notify-icon-inverted;
 
 // Drawer


kirsty-hames (approved):
> From my understanding, hover states do not need to pass accessibility contrast requirements (citation needed...).

That's my understanding too. Colour contrast only applies to adjacent colours, not the differentiation between state colours. 
We're not required to define the boundary of the button however it is recommended to aid people with cognitive disabilities.  

Ref: _Boundaries_ and _Adjacent colours_ https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html

joe-allen-89 (approved):

oliverfoster (approved):

PR #359, opened 2022-11-09
Breaking: block padding property (fixes #358)

### LINKED ISSUE #358: Consolidate block padding custom class variants into property
Within the framework and authoring tool, a user is required to manually type out the custom class(es) required for increasing, decreasing, or removing the block top and / or bottom padding.

This custom class has to be applied to the _classes field on the block or corresponding field within the authoring tool.

### Your environment
* FW / AAT

### Expected behaviour
The application of the block padding custom classes could be optimised by converting the options to a drop down selection, similar to the vanilla's background image styles property.

guywillis:
Fixes: #358

### Breaking
* Consolidates the blocks top and bottom padding into json properties that apply classes on the block depending on property value

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,10 @@ Auto: The background image is displayed in its original size. Cover: Resize the 
 #### **blocks.json**
 >**\_isDividerBlock** (boolean): - Determines whether the CSS class `is-divider-block` will be applied. Acceptable values are `true` and `false`.
 
+>**\_paddingTop** (string): - Changes the value of the blocks top spacing.
+
+>**\_paddingBottom** (string): - Changes the value of the blocks bottom spacing.
+
 Visit the [**Vanilla** wiki](https://github.com/adaptlearning/adapt-contrib-vanilla/wiki) for more information about how to use and manipulate the theme.
 
 ## Structure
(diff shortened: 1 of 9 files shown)


eleanor-heath (approved):

kirsty-hames (approved):
Works as expected (testing via the config in FW build) and agree with the selector name change as it's more intuitive.

Could you include in the example.json please @guywillis?

kirsty-hames (changes requested):
The legacy classes are still in the _example.json_ even though these have been removed from _theme-extras.less_.

Just for consistency with `_componentVerticalAlignment` should we include the supported properties in the description?
e.g. 
```
// Changes the value of the blocks top and bottom spacing.
// Double: Doubles the blocks bottom spacing. Standard: Retains the standard blocks bottom spacing. Half: Halves the blocks bottom spacing. Remove: Removes the blocks bottom spacing. The default value is 'standard'.
```

guywillis:
> The legacy classes are still in the _example.json_ even though these have been removed from _theme-extras.less_.
> 
> Just for consistency with `_componentVerticalAlignment` should we include the supported properties in the description? e.g.
> 
> ```
> // Changes the value of the blocks top and bottom spacing.
> // Double: Doubles the blocks bottom spacing. Standard: Retains the standard blocks bottom spacing. Half: Halves the blocks bottom spacing. Remove: Removes the blocks bottom spacing. The default value is 'standard'.
> ```

Thank you for the review, very much appreciated.

I shall remove the custom classes from the example.json.

I agree with the need for consistency in the documentation. What would be your preference for how this is set up going forwards?

eleanor-heath (approved):

kirsty-hames (approved):

cahirodoherty-learningpool (approved):
👀

PR #370, opened 2022-11-16
Breaking: Quote styling rework (fixes #351)

### LINKED ISSUE #351: Making quotes semantic to encourage inclusive design
Representing quotes with CSS styling alone is not identified by a screen reader. The text will read as any other body text on the page regardless how different the visual styling is. 

For inclusive design, quote content should be wrapped in either `<blockquote>` or `<q>` tags. https://www.w3.org/TR/WCAG20-TECHS/H49.html

`<blockquote>` should be used for long quotations which may require paragraph breaks. 

`<q>` should be used for shorter quotation. 

Currently we don't have any Adapt styling for `<blockquote>` or `<q>` but styles are inherited via the browser. 

blockquote

We have the option to reset/set a base style for these or even use the existing [.pull-quote-body](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/bb314efec4b6736a435efd2ecec50a8fd8b97d2c/less/_defaults/_font-mixins.less#L256) class? I'd be keen to get thoughts on this.

I would avoid using `.pull-quote-title` and `.pull-quote-body` as suggested in [text.less](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/master/less/plugins/adapt-contrib-text/text.less) for the reasoning above. Quotes should be limited to body text only and not used in titles.

guywillis:
Fixes: #351 

### Breaking
* Reworked quote styling within Adapt to provide a semantically correct approach
    * Removed `pull-quote-title` mixin
    * Removed `is-pull-quote` custom class
    * Amended `pull-quote-body` mixin to more generic `quote-body`
    * Added `quote-attribution` mixin
    * Added new variables `font-style` and `letter-spacing` to both `quote-body` and `quote-attribution`
    * Added small overrides to `blockquote` and `figure` html elements
    * Created new custom class `figure-quote` that can be used in any plugin

### DIFF
--- a/less/core/global.less
+++ b/less/core/global.less
@@ -6,3 +6,25 @@ body {
   display: inline-block;
   vertical-align: middle;
 }
+
+// Quote related styling
+// override browser defined margins with rem unit
+blockquote,
+figure {
+  margin: @item-margin * 2;
+}
+
+// styled quote with attribution
+.figure-quote blockquote,
+.figure-quote q {
+  .quote-body;
+}
+
+.figure-quote blockquote {
+  margin: 0;
+}
+
+.figure-quote figcaption {
+  .quote-attribution;
+}
+
--- a/less/_defaults/_font-config.less
+++ b/less/_defaults/_font-config.less
@@ -159,17 +159,19 @@
 
 // Pull quote
 // --------------------------------------------------
-@pull-quote-title-size: 1.25rem;
-@pull-quote-title-family: @title-family;
-@pull-quote-title-weight: @font-weight-regular;
-@pull-quote-title-line-height: @title-line-height;
-@pull-quote-font-style: italic;
-@pull-quote-letter-spacing: normal;
+@quote-body-size: @body-size * 1.5;
+@quote-body-family: @body-family;
+@quote-body-weight: @font-weight-regular;
+@quote-body-line-height: @body-line-height;
+@quote-body-font-style: normal;
+@quote-body-letter-spacing: normal;
 
-@pull-quote-body-size: 1rem;
-@pull-quote-body-family: @body-family;
-@pull-quote-body-weight: @font-weight-bold;
-@pull-quote-body-line-height: @body-line-height;
+@quote-attribution-size: @body-size;
+@quote-attribution-family: @body-family;
+@quote-attribution-weight: @font-weight-regular;
+@quote-attribution-line-height: @body-line-height;
+@quote-attribution-font-style: normal;
+@quote-attribution-letter-spacing: normal;
 
 // Drawer
 // --------------------------------------------------
--- a/less/_defaults/_font-mixins.less
+++ b/less/_defaults/_font-mixins.less
@@ -243,21 +243,24 @@ strong {
   letter-spacing: @item-title-letter-spacing;
 }
 
-.pull-quote-title() {
-  font-size: @pull-quote-title-size;
-  font-family: @pull-quote-title-family;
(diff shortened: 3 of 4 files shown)


guywillis:
Note: Documentation on how to create a semantically correct quote will be coming and part of the vanilla wiki

kirsty-hames (approved):
Looks good thanks @guywillis

less/core/global.less
```diff
@@ -6,3 +6,25 @@ body {
...
+// Quote related styling
+// override browser defined margins with rem unit
+blockquote,
+figure {
+  margin: @item-margin * 2;
+}
+
+// styled quote with attribution
+.figure-quote blockquote,
+.figure-quote q {
```
StuartNicholls, line 19: I'm not sure about including the <q> here as that tag is for inline quotes - brief quotations that do not require paragraph breaks. So I'd imagine the only time it'd be used the text would be the same as the body text around it (not @body-size * 1.5).

guywillis: You're correct, the blockquote and q elements don't need to be specifically called out in the css here. 

Rather confusingly, you can have a myriad of these html tags intertwined with each other in various ways and it can all be used correctly. 

The use of the blockquote and q tags aren't exclusive to the figure element, they can be used outside of this tag with only minor styling amends attached to the blockquote. The introduction of the figure element here will gives us benefits in the form of: 

1. Allowing for semantically correct quotation, be it via blockquote or q, with an author attribution
2. Allowing for the attachment of a custom class for unique styling

StuartNicholls: Yep, I was just referring the styling of the < q > as both could be used in the same document. 

Agree, if using < figcaption > , < figure > should wrap < blockquote > 

As we've lost the < q > tag in the css, perhaps we should name the variable blockquote-XXX rather than just quote-XXX as its just referring to the actual < blockquote > tag. But thats just a minor thing perhaps. 

I'd also say if we are differentiating between a block quote and a true pull quote, actual Pull quotes should be marked up with a < aside > tag around the lot - so maybe sensible to add some css or a selector for that too similar to your 'figure-quote'.

guywillis: I agree that the q tag should inherit the styling it is surrounded by as there could be multiple on a page. The current set up should allow for that. A q tag used outside of the `figure-quote` class will inherit the body styling - a q tag used within will inherit styles from the `quote-body` mixin. 

The only blockquote style within the `figure-quote` class is to remove the double margin indent that comes from using a blockquote tag within a figure tag. I've left the styling open so both blockquote and q can be - it's not limited to purely blockquote.

Agree that something should exist to support a semantically correct implementation of a true pull quote with some refined styling.

StuartNicholls: Yep, I saw all that - all good. Just thought maybe the less variable names could be specific now. But not precious about that.

kirsty-hames (approved):

kirsty-hames (approved):
Looks good thanks @guywillis. Do the custom classes need adding to _example.json_ or will these just go in the wiki?

guywillis:
> Looks good thanks @guywillis. Do the custom classes need adding to _example.json_ or will these just go in the wiki?

The plan is to add them into the [custom classes](https://github.com/adaptlearning/adapt-contrib-vanilla/wiki/Custom-Classes) section of the vanilla wiki

StuartNicholls (approved):

eleanor-heath (approved):

cahirodoherty-learningpool (approved):
👀

PR #394, opened 2023-02-23
New: Add block horizontal alignment option (fixes #393)

### LINKED ISSUE #393: Centering components horizontally
Currently, there is no way to center a component horizontally. This would be useful to have when using a column class (ex. `col-md-8`) to reduce the width of the component.

### Suggestion
After speaking with @guywillis , we decided:

1. Add a new JSON property called `_componentHorizontalAlignment`
2. Create a new CSS class called `align-horz-center`.

swashbuck:
Fixes #393

### New
* Add new classes for `align-horz-center` and `align-horz-right`
* Add new block schema property for `_componentHorizontalAlignment` with the options of `left`, `center`, or `right`

Once approved and merged, this will need added to the Vanilla theme wiki.

### DIFF
--- a/example.json
+++ b/example.json
@@ -96,7 +96,8 @@
     "_isDividerBlock": false,
     "_paddingTop": "",
     "_paddingBottom": "",
-    "_componentVerticalAlignment": ""
+    "_componentVerticalAlignment": "",
+    "_componentHorizontalAlignment": ""
 }
 
 // components.json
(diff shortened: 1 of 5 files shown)


guywillis (approved):
👍

cahirodoherty-learningpool (approved):
👀

eleanor-heath (approved):

Issue #381, opened 2022-12-06
Vanilla Theme Picker Optimisation


guywillis:
### Subject of the issue
The Vanilla's Theme Picker as part of the Authoring Tool could do with some optimisation to be more user friendly. 

Potential task list:
- [ ] Convert theme picker sections into 'legends' that are collapsible
- [ ] Review variable labels
- [ ] Review variable help text
- [ ] Review render location of variables within the theme picker

### Your environment
* v5.x FW & AAT

kirsty-hames:
I can't remember what 'Review render location of variables within the theme picker' was in reference to @guywillis?

guywillis:
It was to review where the variables render within the structure of the page. e.g. does `@background` belong in misc or should it go under Global.

PR #404, opened 2023-03-16
#381 Vanilla Theme Picker Optimisation

zubairslamdien:
Updates #381

### Update
* Updated theme.schema.json to support new font and custom CSS/Less AAT theme picker settings.


taylortom (changes requested):
As there's no planned support for these extra theme settings in the legacy AAT, we should forget adding the settings to the old properties schema

taylortom (approved):

PR #417, opened 2023-04-06
Fix: Add variable for .on-screen-anim() mixin's transition-delay (fixes #416)

### LINKED ISSUE #416: Add variable for .on-screen-anim() mixin's transition-delay
A transition delay parameter was added to animation mixins in #325 I think it would be useful to have a variable that can be easily overridden, similar to `@duration`.

### Environment
- Vanilla theme v9.6.6

swashbuck:
Fixes #416

### Fix
* Adds `@animation-delay` and sets it to a default of 1s.
* Use this new variable in the four default .on-screen-anim() calls.

### DIFF
--- a/less/_defaults/_animations.less
+++ b/less/_defaults/_animations.less
@@ -3,6 +3,7 @@
 @duration: 0.2s;
 @progress-duration: 0.5s;
 
+@animation-delay: 1s;
 @animation-duration: 2000ms;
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

guywillis (approved):
👀

kirsty-hames (approved):
👀

PR #423, opened 2023-05-04
Fix: Add new menu header color variables (fixes #402)

### LINKED ISSUE #402: Box menu color variables
The Box Menu could use some new color variables so that they can be styled independently of other elements. A typical use case would be when using a background image on the menu page where the text colors are unreadable:

## Examples
Current

With menu header colors

## Suggestions:

1. Instead of using `@heading-color`, create a new variable for `@menu-header-title-color`. This would default to `@heading-color`.

https://github.com/adaptlearning/adapt-contrib-vanilla/blob/1a34b8af8d3894d16b2b5e0e4bc0d01da13d4a43/less/plugins/adapt-contrib-boxmenu/boxMenu.less#L44

2. The `.boxmenu__body` element inherits the color from the `body` styles. Create new variable called `@menu-header-body-color`.
3. The `.boxmenu__instruction` element inherits the color from the `body` styles. Create new variable called `@menu-header-instruction-color`.
4. The `.boxmenu__header-inner` does not use a background color. Create new variable called `@menu-header-background-color` that defaults to 'transparent'.

swashbuck:
Fixes #402 

### Fix
* Adds new color variables for the menu header: `@menu-header-background-color`, `@menu-header-title-color`, `@menu-header-body-color`, and `@menu-header-instruction-color`
* Updates schemas

### DIFF
--- a/less/_defaults/_colors.less
+++ b/less/_defaults/_colors.less
@@ -87,6 +87,11 @@
 
 // Menu
 // --------------------------------------------------
+@menu-header-background-color: transparent;
+@menu-header-title-color: @heading-color;
+@menu-header-body-color: @font-color;
+@menu-header-instruction-color: @font-color;
+
 @menu-item: @white;
 @menu-item-inverted: @font-color;
 @menu-item-border-color: @item-color;
--- a/less/plugins/adapt-contrib-boxmenu/boxMenu.less
+++ b/less/plugins/adapt-contrib-boxmenu/boxMenu.less
@@ -36,12 +36,13 @@
   &__header-inner {
     position: relative;
     .l-container-padding(@menu-header-padding-ver, @menu-header-padding-hoz);
+    background-color: @menu-header-background-color;
   }
 
   &__title {
     margin-bottom: @menu-title-margin;
     .menu-title;
-    color: @heading-color;
+    color: @menu-header-title-color;
   }
 
   &__subtitle {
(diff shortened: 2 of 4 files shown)


eleanor-heath (approved):

swashbuck:
@kirsty-hames Thanks for the good feedback. See my thoughts below:

1) I _do_ think that we should create color variables for the page header. I agree that specific color variables are easier to manage in the AAT than needing to remember the exact background classes to use or having to pull in a FED to create custom colors. Happy to include it as part of this PR, or we could create a new ticket for it.
2) We could just add another variable for the "outer" header. `@menu-header-background-color` could be used for the entire `.menu__header` element, and then `@menu-header-inner-background-color` would be used for the header inner. This would give users the most flexibility and doesn't add too much complication.

swashbuck:
> Sounds fine to me but worth mentioning we do have a page header plugin on the horizon adaptlearning/adapt_framework#2703. I'm not sure if the intention is a replacement or alternative to the existing core menu/page headers. Either way, is this something we need to consider when reviewing this PR?

@kirsty-hames Thanks for bringing this up. I think we could still move forward with the menu header and page header color variables and then possibly incorporate them into a new page header plugin (if and when that plugin is released). My assumption is that the current menu and page header implementations will still be supported if people don't want to use the new plugin(s). @oliverfoster Thoughts?

> I'm assuming Vanilla will need to cater for both (plugin and core) so we might need some alignment with use/naming of variables?

Agree, would you want to look at this and confirm the naming conventions for the variables?

oliverfoster:
Yea, carry on working with it as it is, we'll consider the header component later, when it's all ready.

cahirodoherty-learningpool (approved):
👀

PR #429, opened 2023-05-13
Fix: Match columns to breakpoints

### LINKED ISSUE #428: Realign columns to fit new breakpoints
### Subject of the issue
Column classes need to match new breakpoints.

oliverfoster:
fixes #428 

### Fix
* Match columns to breakpoints

### DIFF
--- a/less/project/columns.less
+++ b/less/project/columns.less
@@ -16,8 +16,8 @@
 // min small viewport takes up 8 out of 12 columns
 // min medium viewport takes up 10 columns
 // if undefined element defaults to 100% width
-.make-columns(xs, @device-width-small, 12);
-.make-columns(sm, 400, 12);
+.make-columns(xs, 0, 12); // deprecated: backward compatibility only
+.make-columns(sm, @device-width-small, 12);
 .make-columns(md, @device-width-medium, 12);
 .make-columns(lg, @device-width-large, 12);
 .make-columns(xl, @device-width-xlarge, 12);
--- a/less/project/columns.less
+++ b/less/project/columns.less
@@ -16,8 +16,8 @@
 // min small viewport takes up 8 out of 12 columns
 // min medium viewport takes up 10 columns
 // if undefined element defaults to 100% width
-.make-columns(xs, 0, 12);
-.make-columns(sm, @device-width-small, 12);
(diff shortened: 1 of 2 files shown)


cahirodoherty-learningpool (approved):
👀

less/project/columns.less
```diff
@@ -16,8 +16,8 @@
 // min small viewport takes up 8 out of 12 columns
 // min medium viewport takes up 10 columns
 // if undefined element defaults to 100% width
-.make-columns(xs, 0, 12);
-.make-columns(sm, @device-width-small, 12);
+.make-columns(xs, @device-width-small, 12);
+.make-columns(sm, 400, 12);
```
guywillis, line 20: I'm a little confused as to the changes here - please bear with me in case I have missed something.

To align with the recent breakpoint changes, would something like this may make more 'sense'?

```
.make-columns(sm, 0, 12);
.make-columns(md, @device-width-small, 12);
.make-columns(lg, @device-width-medium, 12);
.make-columns(xl, @device-width-large, 12);
.make-columns(max, @device-width-xlarge, 12);
```

It should follow the same thought process as Adapt's Vanilla styling whereby you target the desired breakpoint and above. 

So if you wanted to apply a custom column style to a component on small devices and above you would use the `sm` column class. This could then be overridden at a wider screen using `lg` to target 960px wide screens and above. 

Diagram this comment is based on - https://github.com/adaptlearning/adapt-contrib-core/issues/364#issuecomment-1544338677

We should try to avoid using an absolute value where possible.

guywillis (approved):
👍

less/project/columns.less
```diff
@@ -16,8 +16,8 @@
 // min small viewport takes up 8 out of 12 columns
 // min medium viewport takes up 10 columns
 // if undefined element defaults to 100% width
-.make-columns(xs, 0, 12);
+.make-columns(xs, 0, 12); // deprecated: backward compatibility only
```
StuartNicholls, line 19: I think maybe XS should stay anyway as its deprecated in the sense that the defaults breakpoint vals make it not needed but I guess it's possible it may be useful in some cases if someone sets the small breakpoint above zero?

StuartNicholls (approved):

Issue #426, opened 2023-05-11
Increase container horizontal padding


guywillis:
### Subject of the enhancement 
Proposal to increase the container horizontal spacing from `.5rem` to `1rem` to allow for more breathing space between the content and the edge of the device.

### Your environment
* Master

StuartNicholls:
@guywillis Does this mean the margins around the content container?

oliverfoster:
The gutter down each side of the content on a phone screen, it's too thin at the moment.

StuartNicholls:
@guywillis / @oliverfoster I've created a pull request #435 that addresses the horizontal space between the gutters & the margins, So desktop layout, increase to the gutters (the space between columns) and the minimum margins (the space either size of the content) in mobile view.

guywillis:
> @guywillis Does this mean the margins around the content container?

Yes, the horizontal container padding attached to elements like blocks and notify

PR #435, opened 2023-05-22
Fix: Increase container horizontal padding (gutters & margins)

StuartNicholls:
Gutters are a little cramped in desktop view as are (layout) margins in mobile view.

fixes: #426

### DIFF
--- a/less/_defaults/_spacing-config.less
+++ b/less/_defaults/_spacing-config.less
@@ -7,7 +7,7 @@
 @item-margin: 0.5rem;
 @item-padding: 1rem;
 
-@item-border-radius: 0.375rem;
+@item-border-radius: 5px;
 
 // Button
 // --------------------------------------------------
@@ -49,7 +49,7 @@
 @block-body-margin: 1rem;
 @block-instruction-margin: 1rem;
 
-@component-title-margin: 1rem; 
+@component-title-margin: 1rem;
 @component-body-margin: 1rem;
 @component-instruction-margin: 1rem;
 
--- a/less/_defaults/_spacing-config.less
+++ b/less/_defaults/_spacing-config.less
@@ -7,7 +7,7 @@
 @item-margin: 0.5rem;
 @item-padding: 1rem;
 
-@item-border-radius: 5px;
+@item-border-radius: 0.375rem;
 
 // Button
 // --------------------------------------------------
@@ -49,7 +49,7 @@
 @block-body-margin: 1rem;
 @block-instruction-margin: 1rem;
 
-@component-title-margin: 1rem;
+@component-title-margin: 1rem; 
 @component-body-margin: 1rem;
 @component-instruction-margin: 1rem;
 
--- a/less/_defaults/_spacing-containers.less
+++ b/less/_defaults/_spacing-containers.less
@@ -21,8 +21,6 @@
 @menu-header-padding-ver: 2rem;
 @menu-header-padding-hoz: 1rem;
 
-@menu-item-container-padding-hoz: 1rem;
-
 @menu-item-padding-ver: 1rem;
 @menu-item-padding-hoz: .5rem;
 
--- a/less/_defaults/_spacing-containers.less
+++ b/less/_defaults/_spacing-containers.less
@@ -43,7 +43,7 @@
 
 // Component padding
 // --------------------------------------------------
-@component-padding-ver: 0.5rem;
+@component-padding-ver: 1rem;
 @component-padding-hoz: 1rem;
 
 // Notify padding
--- a/less/_defaults/_spacing-containers.less
+++ b/less/_defaults/_spacing-containers.less
@@ -21,6 +21,11 @@
 @menu-header-padding-ver: 2rem;
 @menu-header-padding-hoz: 1rem;
 
+@menu-item-container-padding-hoz: 1rem;
+
+@menu-item-padding-ver: 1rem;
+@menu-item-padding-hoz: .5rem;
+
 @menu-item-container-margin: 1.5rem;
 
 // Page container and header padding
--- a/less/_defaults/_spacing-containers.less
+++ b/less/_defaults/_spacing-containers.less
@@ -19,7 +19,7 @@
 @menu-padding-hoz: 0;
 
 @menu-header-padding-ver: 2rem;
-@menu-header-padding-hoz: .5rem;
+@menu-header-padding-hoz: 1rem;
 
 @menu-item-container-margin: 1.5rem;
 
@@ -29,27 +29,27 @@
 @page-padding-hoz: 0;
 
 @page-header-padding-ver: 2rem;
-@page-header-padding-hoz: .5rem;
+@page-header-padding-hoz: 1rem;
 
 // Article header padding
 // --------------------------------------------------
 @article-header-padding-ver: 0;
-@article-header-padding-hoz: .5rem;
+@article-header-padding-hoz: 1rem;
 
 // Block padding
 // --------------------------------------------------
 @block-padding-ver: 4rem;
-@block-padding-hoz: .5rem;
+@block-padding-hoz: 1rem;
 
 // Component padding
 // --------------------------------------------------
-@component-padding-ver: .5rem;
-@component-padding-hoz: .5rem;
+@component-padding-ver: 0.5rem;
+@component-padding-hoz: 1rem;
 
 // Notify padding
 // --------------------------------------------------
 @notify-padding-ver: 4rem;
-@notify-padding-hoz: .5rem;
+@notify-padding-hoz: 1rem;
 
 // Drawer container
 // --------------------------------------------------
--- a/less/plugins/adapt-contrib-boxmenu/boxMenu.less
+++ b/less/plugins/adapt-contrib-boxmenu/boxMenu.less
@@ -65,8 +65,6 @@
   }
 
   &__item-container {
-    margin: 0 -@menu-item-padding-hoz;
-    padding: 0 @menu-item-container-padding-hoz;
     margin-top: @menu-item-container-margin;
   }
 }
--- a/less/plugins/adapt-contrib-boxmenu/boxMenu.less
+++ b/less/plugins/adapt-contrib-boxmenu/boxMenu.less
@@ -65,6 +65,8 @@
   }
 
   &__item-container {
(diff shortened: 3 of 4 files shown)


less/_defaults/_spacing-config.less
```diff
@@ -7,7 +7,7 @@
 @item-margin: 0.5rem;
 @item-padding: 1rem;
 
-@item-border-radius: 5px;
+@item-border-radius: 0.375rem;
```
guywillis, line 10: Can this be reverted please

less/_defaults/_spacing-config.less
```diff
@@ -49,7 +49,7 @@
 @block-body-margin: 1rem;
 @block-instruction-margin: 1rem;
 
-@component-title-margin: 1rem;
+@component-title-margin: 1rem;
```
guywillis, line 52: Please remove white space at end of line

less/_defaults/_spacing-containers.less
```diff
@@ -29,27 +29,27 @@
...
 // --------------------------------------------------
 @block-padding-ver: 4rem;
-@block-padding-hoz: .5rem;
+@block-padding-hoz: 1rem;
 
 // Component padding
 // --------------------------------------------------
-@component-padding-ver: .5rem;
-@component-padding-hoz: .5rem;
+@component-padding-ver: 0.5rem;
```
guywillis, line 46: Can this value be increased to 1rem to ensure component padding is equal in all directions

guywillis (changes requested):

less/_defaults/_spacing-config.less
StuartNicholls: Of course. Can I why this is in px, just not clear to me.

less/_defaults/_spacing-containers.less
StuartNicholls: I did look at this, but increasing to 1rem make it look untidy to my eyes. Will update though

guywillis (approved):

guywillis:
Thanks!

less/_defaults/_spacing-config.less
guywillis: Oversight. You're correct in identifying that the value will need converting to rem to ensure relative scaling works with user defined font sizes.

User font medium, border radius 8px
Screenshot 2023-05-22 at 10 28 50

User font medium, border radius .5rem
Screenshot 2023-05-22 at 10 28 59

Both of the above screenshots look identical as is expected.

User font very large, border radius 8px
Screenshot 2023-05-22 at 10 29 12

User font very large, border radius .5rem
Screenshot 2023-05-22 at 10 29 17

The above two screenshots show a difference in border radius implementation when the user changes their browser font from medium (default) to very large. 

The reason for the revert is that the change sits outside of the remit of this pull request and should be it's own issue where it can be explained.

Tested in Chrome Mac OSX

less/_defaults/_spacing-containers.less
```diff
@@ -19,7 +19,7 @@
 @menu-padding-hoz: 0;
 
 @menu-header-padding-ver: 2rem;
-@menu-header-padding-hoz: .5rem;
+@menu-header-padding-hoz: 1rem;
```
kirsty-hames, line 22: Bumping up the menu header means this doesn't align with the menu items below. Not sure if we would want to bump the [menu item margin](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/aac3a2bdade9398bc9cf353dd40b64f5235096dd/less/plugins/adapt-contrib-boxmenu/boxMenuItem.less#L3) for consistency?
header_and_item_alignment

guywillis: Agreed

kirsty-hames: Although, I think bumping the menu item margin could be more problematic with scaling/overrun issues. Particularly if you're using column classes etc where your items are rendered at smaller widths. Maybe revert menu header back to `0.5rem` rather than amending the menu item margin? Happy to go with whichever people feel is less of an issue.

StuartNicholls: The menu item margin in hard-coded in the theme. I think the theme needs an overall modernisation and rationalisation. I'll create a separate issue for this, I just need to do a bit more prep work. In the meantime lest try and tidy these bits up. I'll have a at any issues this may cause now.

StuartNicholls: @kirsty-hames, we can't increate the item spacing to effect the layout margins as that increases the gutter between the items. Suggest adding a 0.5 rem margin to .boxmenu__item-container to compensate for now. Then we can do a better review holistically when I create a new issue. 

@guywillis / @kirsty-hames - what do you think?

kirsty-hames: > @kirsty-hames, we can't increase the item spacing to affect the layout margins as that increases the gutter between the items. Suggest adding a 0.5 rem margin to .boxmenu__item-container to compensate for now. Then we can do a better review holistically when I create a new issue.
>
Yep I'd rather not increase the gutter. I'm happy to go with your suggestion thanks.

StuartNicholls: There isn't really a perfect way of doing this currently without a bit of a rethink. In my opinion, the best solution is to add an extra var -> @menu-item-container-padding-hoz: 1rem;
 then update that item container with this:
 
```css
   &__item-container {
    margin: 0 -@menu-item-padding-hoz;
    padding: 0 @menu-item-container-padding-hoz;
  }
```
 
 I'll also add extra vars of these and update box menu also:
 
@menu-item-padding-ver: 1rem;
@menu-item-padding-hoz: 0.5rem;

Obviously, boxmenu would require these updates to the theme. Is there a formal way where we document this?

less/plugins/adapt-contrib-boxmenu/boxMenu.less
```diff
@@ -65,6 +65,8 @@
   }
 
   &__item-container {
+    margin: 0 -@menu-item-padding-hoz;
```
kirsty-hames, line 68: The negative margin is causing a vertical scrollbar to display. Testing on Mac Safari, FF, Chrome and iPhone.

less/plugins/adapt-contrib-boxmenu/boxMenuItem.less
```diff
@@ -1,6 +1,6 @@
 .boxmenu-item {
   &__inner {
-    margin: 0 0.5rem 1rem;
+    margin: 0 @menu-item-padding-hoz @menu-item-padding-ver;
```
kirsty-hames, line 3: Although variables offer flexibility over hard coded I'm not sure how often these get amended. As the boxmenu doesn't support a proper grid or column layout currently I think these values are pretty set.
I'd also expect padding variables to apply to the inner spacing rather than the gutter.

kirsty-hames (commented):
> There isn't really a perfect way of doing this currently without a bit of a rethink.

I agree, I think the column/grid layout needs some consideration as we're compensating for how guttering is handled by the item. Rather than investing too much time on trying to work around a bigger issue would reverting the `@menu-header-padding-hoz` back to `0.5rem` be the simpler solution for now?

less/plugins/adapt-contrib-boxmenu/boxMenu.less
StuartNicholls: Well, that's inconvenient. The nag margins is a bit of an old skool way of doing things now we have gap and flex. It really should be in an additional div that wraps the items. So, I think we have three options for the short term, 1. leave with a smaller margin on the item container / 2. try to fix with additional div etc. or 3. revert

Adjusting a margin should be really easy, but I in some ways at least it highlights what needs addressing . I'm happy to go all 3 routes though.

less/plugins/adapt-contrib-boxmenu/boxMenuItem.less
kirsty-hames: I'm happy for these to remain.

less/plugins/adapt-contrib-boxmenu/boxMenu.less
kirsty-hames: I'd be in favour of option 3 (revert) and instead invest effort looking into improving the column/grid layout as a separate task. However, if people want to see a quick fix for this then I'm happy with that too.

StuartNicholls: also happy either way, @guywillis ?

guywillis: Am in agreement with Kirsty here. The styles need fixing but it's probably best to invest the time into a long term solution.

StuartNicholls: So revert or live with the slight inconsistency ? @guywillis / @kirsty-hames

StuartNicholls: Sorry, definitely remove the neg margins. I meant reverting the margin increase so it was consistent (so for the page also). But can just revert for menu.. Or I was also suggesting we could leave the menu header, but just have the inconsistency in the item container.

StuartNicholls: @kirsty-hames / @guywillis I've reverted the negative margins on the menu items. This does mean that everything apart from the margin around the items is 1rem and here its 0.5 but I've created a new issue to address this here #447

less/_defaults/_spacing-containers.less
```diff
@@ -19,7 +19,12 @@
...
+
+@menu-item-container-padding-hoz: 1rem;
```
kirsty-hames, line 24: Is `@menu-item-container-padding-hoz` needed @StuartNicholls?

kirsty-hames (commented):
I'm happy to accept the current margin differentiation on the menu and instead focus efforts on #447 rather than bodge job the current issue.

I've raised a query on `@menu-item-container-padding-hoz` otherwise it's an approval from me.

swashbuck (approved):
👍

less/_defaults/_spacing-containers.less
StuartNicholls: > Is `@menu-item-container-padding-hoz` needed @StuartNicholls?

Nope, added that in when thinking I could fix the menu items. Removing it for now

PR #448, opened 2023-06-15
Update: replace notify button icon duplicate styles with mixin (fixes #443)

### LINKED ISSUE #443: Reduce duplicate code across plugins using notify popup
Currently the default style for Notify icon buttons is consistent across all plugins however we've duplicated the styling code across the various plugin LESS files. 

See Hotgraphic close button in comparison to core Notify close button. The CSS being applied is the same but attached to different selectors.

notify_close_btn

See references to duplicate styles below:
- Notify targets [`.notify__btn-icon`](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/142b4fde60d638e280d04d691c63fbe62bd5abfe/less/core/notify.less#LL74C1-L86C4)
- Tutor targets [`.tutor__btn-icon`](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/142b4fde60d638e280d04d691c63fbe62bd5abfe/less/plugins/adapt-contrib-tutor/tutor.less#L20)
- Hotgraphic targets [`.hotgraphic-popup__controls` and `.hotgraphic-popup__close`](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/142b4fde60d638e280d04d691c63fbe62bd5abfe/less/plugins/adapt-contrib-hotgraphic/hotgraphicPopup.less#LL9C1-L22C4)

### Expected behaviour
The generic Notify button icon selector `.notify__btn-icon` is only being used in the [notifyPopup](https://github.com/adaptlearning/adapt-contrib-core/blob/24643de7fd2c62adca3ed2bff19ef48be460d73d/templates/notifyPopup.hbs#L98) template.

I'd expect `.notify__btn-icon` to be used for all Notify icon buttons. This means the default style is set in one place but we still have plugin specific classes to target should you want to style these different - although there's good reason for keeping UI button styling consistent. 

### Next steps
- Raise issues and PRs for the plugins where the additional selector `.notify__btn-icon` is applicable.
- Only once plugin PRs are merged should the plugin styles be removed from Vanilla. As we're removing styles that require the compatible plugin? Is this a breaking change for Vanilla or the relevant plugin? Or both?

kirsty-hames:
Fixes https://github.com/adaptlearning/adapt-contrib-vanilla/issues/443

The default style for Notify icon buttons is consistent across all plugins however the styles are duplicated across the various plugin LESS files.

Vanilla instances include:
- Notify popup and Notify push
- Hotgraphic popup
- Tutor popup

This PR wraps the existing icon button styles within a mixin and applies the mixin in the various plugin LESS files to replace duplicated code.

As the mixin is specific to Notify, the mixin remains in notify.less.

### DIFF
--- a/less/plugins/adapt-contrib-hotgraphic/hotgraphicPopup.less
+++ b/less/plugins/adapt-contrib-hotgraphic/hotgraphicPopup.less
@@ -8,17 +8,7 @@
 
   &__controls,
   &__close {
-    margin: @btn-margin / 2;
-    padding: @btn-padding / 2;
-    background-color: @notify-icon;
-    color: @notify-icon-inverted;
-    border-radius: 50%;
-
-    .no-touch &:not(.is-disabled):not(.is-locked):hover {
-      background-color: @notify-icon-hover;
-      color: @notify-icon-inverted-hover;
-      .transition(background-color @duration ease-in, color @duration ease-in;);
-    }
+    .notify-btn-icon;
   }
 
   &__item-title {
--- a/less/core/notify.less
+++ b/less/core/notify.less
@@ -72,17 +72,7 @@
   }
 
   &__btn-icon {
-    margin: @btn-margin / 2;
-    padding: @btn-padding / 2;
-    background-color: @notify-icon;
-    color: @notify-icon-inverted;
-    border-radius: 50%;
-
-    .no-touch &:not(.is-disabled):not(.is-locked):hover {
-      background-color: @notify-icon-hover;
-      color: @notify-icon-inverted-hover;
-      .transition(background-color @duration ease-in, color @duration ease-in;);
-    }
+    .notify-btn-icon;
   }
 
   &__close-btn {
@@ -108,4 +98,20 @@
   //     left: 0;
   //   }
   // }
+}
+
+// Global Notify button icon
+// --------------------------------------------------
+.notify-btn-icon() {
+  margin: @btn-margin / 2;
+  padding: @btn-padding / 2;
+  background-color: @notify-icon;
+  color: @notify-icon-inverted;
+  border-radius: 50%;
+
+  .no-touch &:not(.is-disabled):not(.is-locked):hover {
+    background-color: @notify-icon-hover;
+    color: @notify-icon-inverted-hover;
+    .transition(background-color @duration ease-in, color @duration ease-in;);
+  }
 }
--- a/less/core/notifyPush.less
+++ b/less/core/notifyPush.less
@@ -37,17 +37,7 @@
     position: absolute;
     top: 0;
     right: 0;
-    margin: @btn-margin / 2;
-    padding: @btn-padding / 2;
-    background-color: @notify-icon;
-    color: @notify-icon-inverted;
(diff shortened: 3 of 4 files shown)


swashbuck (approved):
👍

cahirodoherty-learningpool (approved):
👀

less/plugins/adapt-contrib-hotgraphic/hotgraphicPopup.less
```diff
@@ -8,17 +8,7 @@
...
-    color: @notify-icon-inverted;
-    border-radius: 50%;
-
-    .no-touch &:not(.is-disabled):not(.is-locked):hover {
-      background-color: @notify-icon-hover;
-      color: @notify-icon-inverted-hover;
-      .transition(background-color @duration ease-in, color @duration ease-in;);
-    }
+    .notify-btn-icon;
   }
```
StuartNicholls, line 12: Some small things, perhaps for another PR?

I'm not sure about using the same mixin for both the controls and the close button. I'd imagine these may want to be styled differently. For example, good UI/UX might say you don't really want to give the close button and the nav buttons equal prominence. Suggested [here](https://github.com/adaptlearning/adapt-contrib-core/issues/386)

Additionally, the close button isn't ever disabled/locked so thats also a bit confusing. 

Note: We also might want to make the border-radius a less variable? (defo a new PR)

kirsty-hames: Valid points thanks @StuartNicholls and I agree. As the current styling of controls and close buttons uses the same variables and styles I've maintained this as the purpose of this PR was to just remove the duplication. I think border-radius and controls button styling can both be raised as new issues/PRs.

StuartNicholls: Cool, the only thing is, do we need a different target in the templates? Fine for a separate PR here but maybe need to rationalise the classes in templates. I'll put a note [here](https://github.com/cgkineo/adapt-hotgrid/pull/127)

StuartNicholls (approved):

less/plugins/adapt-contrib-hotgraphic/hotgraphicPopup.less
kirsty-hames: Just picking this back up again. I'll merge this PR as it satisfies the [issue](https://github.com/adaptlearning/adapt-contrib-vanilla/issues/443) raised, to reduce duplicate code across plugins. I think the points made in this discussion can be tackled in separate PRs @StuartNicholls.

> I'm not sure about using the same mixin for both the controls and the close button. I'd imagine these may want to be styled differently. For example, good UI/UX might say you don't really want to give the close button and the nav buttons equal prominence. Suggested https://github.com/adaptlearning/adapt-contrib-core/issues/386

> Additionally, the close button isn't ever disabled/locked so thats also a bit confusing.

I think both of these will be part of the wider Button review in https://github.com/adaptlearning/adapt-contrib-core/issues/386

> We also might want to make the border-radius a less variable?

I'll raise a new issue off the back of this.

PR #454, opened 2023-07-07
Fix: Remove background color on Hot Graphic pins when using pin images (fixes #453)

### LINKED ISSUE #453: Hot Graphic background color when using graphics as pins
### Subject of the issue
When using graphics as pins, the `button.hotgraphic__pin` element inherits the background color from `.btn-icon`.  As such, you can see the button's background color when using transparent PNGs.

In this example, standard buttons have a gray background.  This color is visible behind the star pin graphics:

hotgraphic-pin

### Your environment
* Framework 5.24.8
* Vanilla theme 9.3.1
* Hot Graphic 6.5.2

### Steps to reproduce
1. Add a Hotgraphic.
2. Configure to use a graphic for a pin with the`_pin` option:

```
"_pin": {
    "src": "course/en/images/hotgraphic-pin.png",
    "alt": ""
}
```

### Suggested next steps
Couple ideas for this:
1. (preferred) Always use a transparent background for graphic pins.
```
.hotgraphic__pin.has-pin-image {
  background-color: transparent;
}
```
2. Add a new setting called `_isTransparent` that would live in the `_pin` object.  When this is true, the background color would be set to transparent.

swashbuck:
Fix #453 

### Fix
* Remove background color on Hot Graphic pins when using pin images

### Testing
See #453

### DIFF
--- a/less/plugins/adapt-contrib-hotgraphic/hotgraphic.less
+++ b/less/plugins/adapt-contrib-hotgraphic/hotgraphic.less
@@ -1,4 +1,12 @@
 .hotgraphic {
+  &__pin.has-pin-image {
+    background-color: transparent;
+
+    .no-touch &:not(.is-disabled):not(.is-locked):hover {
+      background-color: transparent;
+    }
+  }
+
   &__pin:not(.has-pin-image) {
     background-color: @item-color;
     color: @item-color-inverted;


cahirodoherty-learningpool (approved):
👀

kirsty-hames (approved):
The current `background-color` is inherited from the `.btn-icon` class which I think should only get rendered for icon pins, not image pins. Regardless of the class, the `background-colour` would inherit from the browser `<button>` style anyway so you would still need to override this (as per your PR).

This works for me thanks @swashbuck.

oliverfoster (approved):

PR #458, opened 2023-08-01
Fix: Add 'page-view-ready' html class once page is ready

### LINKED ISSUE #457: Add html class after all assets are loaded for a page
To help with the timing of CSS animations, it would be useful to have an `html` class that indicates when all assets have been loaded for a page.

### Use case
I want to animate the title element as soon as the page is loaded and the loading screen is removed. There is currently not a way to know when the loading screen is removed, so it's difficult to know when the animation should be triggered. For instance, there is a chance that the animation is triggered while the loading screen is still overlapping the content, so the animation will never be seen. With a new class like `page-view-ready`, I can do the following:

```
.page__title {
  transition: 0.5s all ease-in-out;
}
html.page-view-ready .page__title {
  color: fuchsia;
  transform: scale(1.05) rotate(-1deg) translate(1rem, -1rem);
}
```

https://github.com/adaptlearning/adapt-contrib-vanilla/assets/898168/4b36dc74-f705-4b71-8268-607a1be1ee04

swashbuck:
Fixes adaptlearning/adapt-contrib-core#423 

### Fix
* Adds the `page-view-ready` html class once the page is ready

### Testing
Add some CSS that animates an element once the page is ready and the loading screen has been removed.

You may want to enable device throttling or add a large image to the page to increase the loading time.

```
.page__title {
  transition: 0.5s all ease-in-out;
}

html.page-view-ready .page__title {
  color: fuchsia;
  transform: scale(1.05) rotate(-1deg) translate(1rem, -1rem);
}
```
The animation should not play until the loading screen has been removed.

### DIFF
--- a/js/themeView.js
+++ b/js/themeView.js
@@ -6,8 +6,10 @@ export default class ThemeView extends Backbone.View {
 
   initialize() {
     this.setStyles();
+    this.resetPageViewReady();
 
     this.listenTo(Adapt, {
+      'pageView:ready': this.onPageViewReady,
       'device:changed': this.onDeviceResize,
       remove: this.remove
     });
@@ -15,6 +17,15 @@ export default class ThemeView extends Backbone.View {
 
   onDeviceResize() {
     this.setStyles();
+  }
+
+  onPageViewReady() {
+    // Add a class to <html> after all assets are loaded for a page.
+    $('html').addClass('page-view-ready');
+  }
+
+  resetPageViewReady() {
+    $('html').removeClass('page-view-ready');
   }
 
   remove() {


eleanor-heath (approved):

oliverfoster:
Can you add this [here](https://github.com/adaptlearning/adapt-contrib-core/blob/3f9b44d1c1dedcaf70adc75073982827f588d3b0/js/views/contentObjectView.js#L76) instead please?
Make it `is-ready` instead?

swashbuck:
Thanks, @oliverfoster . I'll close this PR and will move the ticket to the `adapt-contrib-core` space.

PR #459, opened 2023-08-11
Fix: Add schema default values for background image styles (fixes #456)

### LINKED ISSUE #456: AAT: Default values for background image styles
### Subject of the issue
Currently in the AAT, there are no default values for Background Image Styles. Instead, the dropdown menus are initially set to nothing.

Although there are default styles within the Less/CSS, I think it would be less ambiguous for course creators if the default values were already populated in the form fields. Otherwise, you leave users scratching their heads about whether they need to choose values for the fields.

I would like to see the default Less/CSS values applied to the schemas for the following fields:

- **_backgroundRepeat:** no-repeat
- **_backgroundSize:** cover
- **_backgroundPosition:** center top

In general, I think all dropdowns should have a default value, so we could also do this for Text alignment and any other fields without defaults.

**Question:** Is there a reason why these defaults are not set? If not, I can throw a PR together for this.

swashbuck:
Fix #456 

### Fix
* Adds schema default values for Background image styles

### Testing
1. Upload plugin to an AAT instance
2. Verify that there are default values for Repeat, Size and Position under Background image styles

### DIFF
--- a/schema/article.schema.json
+++ b/schema/article.schema.json
@@ -117,7 +117,6 @@
                   "description": "Repeat-x: The background image is repeated only horizontally. Repeat-y: The background image is repeated only vertically",
                   "default": "no-repeat",
                   "enum": [
-                    "",
                     "repeat",
                     "repeat-x",
                     "repeat-y",
@@ -131,7 +130,6 @@
                   "description": "Auto: The background image is displayed in its original size. Cover: Resize the background image to cover the entire container, even if it has to stretch or crop the image. Contain: Resize the background image to make sure the image is fully visible. 100% 100%: Resize the background image to match the dimensions of the container",
                   "default": "cover",
                   "enum": [
-                    "",
                     "auto",
                     "cover",
                     "contain",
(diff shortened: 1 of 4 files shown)


schema/article.schema.json
```diff
@@ -115,7 +115,7 @@
                   "type": "string",
                   "title": "Repeat",
                   "description": "Repeat-x: The background image is repeated only horizontally. Repeat-y: The background image is repeated only vertically",
-                  "default": "",
+                  "default": "no-repeat",
                   "enum": [
                     "",
```
guywillis, line 120: We should probably remove the empty string `""` from the list of options for article / block / content object schemas if we're setting a default

eleanor-heath (approved):

schema/article.schema.json
swashbuck: Agreed and removed

oliverfoster (approved):

kirsty-hames (approved):

PR #461, opened 2023-08-18
Fix: Add 'bg-color white' mixin (fixes #460)

### LINKED ISSUE #460: Add bg-color mixin for white background / black text
### Subject of the issue
In `less/project/theme-common.less`, there is not a predefined mixin for white background and black text. This is a mixin that we use often, so it would be useful if it was included by default.

Additional justification: there _is_ a predefined mixin for black background with white text. Just not the reverse.

### Your environment
* Vanilla 9.8.2

swashbuck:
Fix #460 

### Fix
* Adds a white bg-color-mixin() to the out-of-the-box background color mixins

### Testing
1. Add `bg-color white` to a block, component, etc.

### DIFF
--- a/less/project/theme-common.less
+++ b/less/project/theme-common.less
@@ -8,6 +8,7 @@
 // Note: both arguments must be predefined variables
 // --------------------------------------------------
 .bg-color-mixin(black, white);
+.bg-color-mixin(white, black);
 .bg-color-mixin(background, background-inverted);
 .bg-color-mixin(transparent-light, black);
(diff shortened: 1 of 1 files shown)


cahirodoherty-learningpool (approved):
👀

guywillis (approved):
👀

eleanor-heath (approved):

kirsty-hames:
wiki updated https://github.com/adaptlearning/adapt-contrib-vanilla/wiki/Custom-Classes#global-common-classes

PR #463, opened 2023-08-22
Fix: background and headerBackground undefined variable errors (fixes #462)

### LINKED ISSUE #462: Errors when background layer is created by another plugin
### Subject of the issue
Currently, the "background" div layer is [created](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/6679f9fddc0ae8f4d766f4c0972ddf7dd73c4e96/js/themeView.js#L49) by the Vanilla theme in `addBackgroundLayer()`. In some cases, the "background" layer div may be created by another plugin which causes `addBackgroundLayer()` to exit early without setting `this.$background`. The methods `setBackgroundImage()` and `setBackgroundStyles()` then try to reference this property and throw an error when it is not found.

I would suggest that we simply check for this.$background and exit early from those two methods if it's not found.

### Your environment
* Vanilla 9.8.1

### Screenshots
themeView

swashbuck:
Fixes #462

### Fix
* Exit early from `setHeaderBackgroundImage()` and `setHeaderBackgroundStyles()` if `this.$headerBackground` has not been set
* Exit early from `setBackgroundImage()` and `setBackgroundStyles()` if `this.$background` has not been set
* Fix deprecated `Adapt.device` references

### DIFF
--- a/js/themePageView.js
+++ b/js/themePageView.js
@@ -37,7 +37,8 @@ export default class ThemePageView extends ThemeView {
 
   setHeaderBackgroundImage(config, $header) {
     const backgroundImages = config._backgroundImage;
-    if (!backgroundImages) return;
+    if (!backgroundImages || this.$headerBackground) return;
+
     const backgroundImage = backgroundImages[`_${Adapt.device.screenSize}`] ?? backgroundImages._small;
     $header.toggleClass('has-bg-image', Boolean(backgroundImage));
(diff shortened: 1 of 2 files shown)


cahirodoherty-learningpool (approved):
👀

eleanor-heath (approved):

kirsty-hames (approved):
Works as expected thanks @swashbuck

Issue #465, opened 2023-09-13
Is the Boxmenu item button margin needed?


kirsty-hames:
### Subject of the issue
Minor point, but, the `margin-bottom` attached to the `.boxmenu-item__button-container` creates inconsistent spacing around the item (`1rem` top/left/right and `2rem` bottom). See screen shot below.

In addition, the `.boxmenu-item__button-container` margin is set by the [`@menu-item-progress-margin`](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/f85ef8544c262e3a7f25dd1681223a5641e039e6/less/plugins/adapt-contrib-boxmenu/boxMenuItem.less#L41) variable - not sure if this was an oversight when added.

<img width="1489" alt="inconsistent_spacing" src="https://github.com/adaptlearning/adapt-contrib-boxMenu/assets/7045330/47591194-a03f-4b38-8c02-deb50bfb9d2d">

### Expected behaviour
As the `.boxmenu-item__button-container` is the last item on the menu item and there is already padding applied to the `.boxmenu-item__inner`, I don't see having margin here adds any value. This is something I typically remove when styling menus anyway.

kirsty-hames:
Fixed by https://github.com/adaptlearning/adapt-contrib-vanilla/pull/467

PR #467, opened 2023-09-14
Update: use display flex to set equal menu item heights (fixes #466)

### LINKED ISSUE #466: Menu item heights should be even
### Subject of the issue
Currently, the height of menu items is based on the amount of text content within. The visual harmony would be enhanced if the item boxes were always the same height regardless of the contents. This is something I always fix in projects when I'm working with a custom theme.

We could do this fairly easily with CSS Flexbox:

```
.boxmenu-item {
  display: flex;
}
.boxmenu-item__inner {
  width: 100%;
}
```

### Your environment
* Box Menu 6.3.11

### Steps to reproduce
Add different amounts of body text to different content objects

### Screenshots
**Current layout**

**Preferred (equal height boxes)**

**Even better (align duration, progress bar, and "view" button to the bottom)**

kirsty-hames:
Fixes https://github.com/adaptlearning/adapt-contrib-vanilla/issues/466 and https://github.com/adaptlearning/adapt-contrib-vanilla/issues/465

## Set equal menu item heights
Currently, the height of menu items is based on the amount of text content which leads to inconsistent item heights.

This PR uses `display: flex` to display all items at equal height with no constraints on the content. Please see below for screenshots of how this looks in different use cases.

 **Standard display:**
standard_display

**Items using [column layout](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/master/less/project/columns.less):**
column_layout

**Items without image:**
no_image

## Remove button container margin

I've also included #465 in this PR. The `margin-bottom` attached to the `.boxmenu-item__button-container` creates inconsistent spacing around the item (1rem top/left/right and 2rem bottom). As the `.boxmenu-item__button-container` is the last item in the menu item, and there is already padding applied to the `.boxmenu-item__inner`, the margin here isn't needed.

### DIFF
--- a/less/plugins/adapt-contrib-boxmenu/boxMenuItem.less
+++ b/less/plugins/adapt-contrib-boxmenu/boxMenuItem.less
@@ -1,5 +1,8 @@
 .boxmenu-item {
+  display: flex;
+  
   &__inner {
+    width: 100%;
     margin: 0 @menu-item-padding-hoz @menu-item-padding-ver;
     padding: 1rem;
     background-color: @menu-item;
--- a/less/plugins/adapt-contrib-boxmenu/boxMenuItem.less
+++ b/less/plugins/adapt-contrib-boxmenu/boxMenuItem.less
@@ -41,7 +41,6 @@
   &__button-container {
     display: flex;
     align-items: center;
-    margin-bottom: @menu-item-progress-margin;
   }
 
   &__button {


swashbuck (approved):
👍

cahirodoherty-learningpool (approved):
👀

danielghost (approved):

PR #468, opened 2023-09-21
Update: remove default color hex codes from schemas (fixes #464)

### LINKED ISSUE #464: schema color variables aren't consistent with LESS color variables
### Subject of the issue
In a previous PR, the [blue colour variable was updated in the LESS to meet AA color contrast](https://github.com/adaptlearning/adapt-contrib-vanilla/pull/328) but the hex codes referencing the blue in the schemas were not updated.

In the AAT, the schema color variables will always take precedence over the LESS variables so the old blue, `#36CDE8` (pre-PR) still applies and fails colour contrast.

### Expected behaviour
I'd suggest removing the `default` hex codes from the schemas as the default Vanilla colours are already applied via the CSS. Keeping the `default` values is another maintenance task when updating the colour variables - both maintaining Vanilla and creating bespoke themes. As per this issue, it's easily forgotten to update the schemas when amending the LESS variables.

kirsty-hames:
Fixes https://github.com/adaptlearning/adapt-contrib-vanilla/issues/464

The default Vanilla colours are already applied via the CSS. Having schema defaults is another maintenance task when updating the colour variables - both maintaining Vanilla and creating bespoke themes.

### DIFF
--- a/properties.schema
+++ b/properties.schema
@@ -15,14 +15,14 @@
             "title": "Font colour",
             "type": "string",
             "inputType": "ColourPicker",
-            "default": "#4D4D4D",
+            "default": "",
             "help": "Global body font colour."
           },
           "font-color-inverted": {
             "title": "Font colour inverted",
             "type": "string",
             "inputType": "ColourPicker",
-            "default": "#FFFFFF",
+            "default": "",
             "help": "Global body font colour inverted. Either a lightened/darkened alternative to Font colour."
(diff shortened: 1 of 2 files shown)


guywillis (approved):
👀

cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

PR #471, opened 2023-10-02
Update: Update visited color variables (fixes #470)

### LINKED ISSUE #470: Update visited color?
We current use grey for component visited states. Seeing as grey is widely recognised, and used in Adapt for disabled states. Should we be using an alternative colour to differentiate visited from disabled states?

current_visited_state

### Expected behaviour
[`@visited`](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/a524d99e3d104441c5161287d2565fdcf3e11e42/less/_defaults/_colors.less#L50) is only used by presentation components (Accordion, Narrative, Hotgraphic). Therefore I think using `@item-color` (or a percentage of) as per other component states would be more appropriate. For example:
```
@visited: darken(@item-color, 15%);
@visited-inverted: @item-color-inverted;
```
suggested_visited_state

kirsty-hames:
Grey is widely recognised, and used in Adapt for disabled states so changing the visited colours to differentiate between the two. Variables replaced with a percentage of `@item-color` for consistency with other component states.

Fixes https://github.com/adaptlearning/adapt-contrib-vanilla/issues/470

### DIFF
--- a/less/_defaults/_colors.less
+++ b/less/_defaults/_colors.less
@@ -47,8 +47,8 @@
 
 // Visited - presentation components
 // --------------------------------------------------
-@visited: #727272;
-@visited-inverted: @white;
+@visited: darken(@item-color, 15%);
+@visited-inverted: @item-color-inverted;
 
 // Buttons
 // --------------------------------------------------


swashbuck (approved):
👍

oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

PR #474, opened 2023-10-09
Update: Nav styling amends (fixes #472)

### LINKED ISSUE #472: Change navigation visual style
### Subject of the issue
A fairly regular piece of feedback that we receive is that the navigation buttons in Adapt do not look like navigation buttons. 

### Your environment
* Master AT & FW

### Expected behaviour
After some musing and experimentation I wonder whether introducing a small amount of spacing around the navigation buttons themselves goes some way, at least in the short term, to emphasising that the navigation icons and text are actually clickable buttons.

Mock for example:

Screenshot 2023-10-05 at 10 40 10

Comments very much welcomed.

guywillis:
Fixes: https://github.com/adaptlearning/adapt-contrib-vanilla/issues/472

### Update
* Introduced spacing to navigation buttons and navigation inner elements to emphasise that the buttons are clickable items.

Screenshot 2023-10-09 at 14 23 47

### DIFF
--- a/less/core/nav.less
+++ b/less/core/nav.less
@@ -5,9 +5,16 @@
     max-width: inherit;
   }
 
+  &__inner {
+    padding-left: @nav-btn-margin / 2;
+    padding-right: @nav-btn-margin / 2;
+  }
+
   &__btn {
+    margin: @nav-btn-margin (@nav-btn-margin / 2);
     background-color: @nav-icon;
     color: @nav-icon-inverted;
+    border-radius: @nav-btn-border-radius;
 
     .no-touch &:not(.is-disabled):not(.is-locked):hover {
       background-color: @nav-icon-hover;
--- a/less/_defaults/_spacing-config.less
+++ b/less/_defaults/_spacing-config.less
@@ -16,6 +16,12 @@
 
 @btn-border-radius: 0.25rem;
 
+// Navigation
+// --------------------------------------------------
+@nav-btn-margin: .5rem;
+
+@nav-btn-border-radius: @btn-border-radius;
+
 // Menu header text margin
 // --------------------------------------------------
 @menu-title-margin: 1rem;


cahirodoherty-learningpool (approved):
👀

less/core/nav.less
```diff
@@ -5,9 +5,16 @@
     max-width: inherit;
   }
 
+  &__inner {
+    padding-left: @nav-btn-margin / 2;
+    padding-right: @nav-btn-margin / 2;
+  }
+
   &__btn {
+    margin: @nav-btn-margin (@nav-btn-margin / 2);
```
eleanor-heath, line 14: did you mean `margin: calc(@nav-btn-margin / 2);` here?

less/core/nav.less
```diff
@@ -5,9 +5,16 @@
     max-width: inherit;
   }
 
+  &__inner {
+    padding-left: @nav-btn-margin / 2;
+    padding-right: @nav-btn-margin / 2;
```
eleanor-heath, line 10:
```
padding-left: calc(@nav-btn-margin / 2);
padding-right: calc(@nav-btn-margin / 2);
```

kirsty-hames (approved):
Looks good and works as expected thanks @guywillis

less/core/nav.less
guywillis: The `@nav-btn-margin` isn't a CSS variable which negates it needing the var prefix. 

The calc function also isn't required as the calculation is fairly basic (we're dividing an absolute number by an absolute number rather than say dividing a dynamic number (%) by an absolute number).

eleanor-heath: @nav-btn-margin (@nav-btn-margin / 2); looked strange to me just, are we multiplying @nav-btn-margin by (@nav-btn-margin divided by 2)?

eleanor-heath: if we are only dividing the CSS variable by 2 should it just be `margin: (@nav-btn-margin / 2)`?

eleanor-heath (approved):

less/core/nav.less
guywillis: We're setting the vertical margin to match the value of `@nav-btn-margin` whilst the horizontal margin equals half of `@nav-btn-margin`.

Admittedly, it does look odd to have differing margins for the vertical and horizontal axis. I have added additional padding onto the `nav__inner` element to account for the reduced spacing to the left of the first element in the nav and to the right of the last as this was a cleaner method than trying to detect first and last visual elements.

eleanor-heath: No problem, I realise now what this is doing! Apologies for initial the confusion.

PR #481, opened 2023-12-15
Update: Updated RTL styling (fixes #480)

### LINKED ISSUE #480: Optimise RTL styling
The RTL styling present in the `button.less`, `drawerItem.less`, `notify.less`, `notifyPush.less`, `accordion.less`, `glossary.less`, `dropdown.less`, `mcq.less`, `narrative.less`, `textinput.less`, `tutor.less`, and `theme-common.less` files could be optimised a touch better by utilising the [margin-inline](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline), [padding-inline](https://developer.mozilla.org/en-US/docs/Web/CSS/padding-inline) properties and [text-align: start | end](https://developer.mozilla.org/en-US/docs/Web/CSS/text-align) instead of the current approach.

The `margin-inline` and `padding-inline` properties and `text-align: start` and `text-align: end` values do not require an associated RTL version as the property inherits the body direction.

Not ie11 compatible:
* https://caniuse.com/?search=margin-inline
* https://caniuse.com/?search=text-align%20start

### Your environment
* Master

guywillis:
Fixes #480 

### Update
* Updated RTL styling to remove unnecessary declarations

### DIFF
--- a/less/core/button.less
+++ b/less/core/button.less
@@ -50,11 +50,6 @@
 
   .can-show-marking &__action.is-full-width,
   .can-show-marking &__feedback.is-full-width {
-    margin-right: @icon-size + (@item-margin * 2) + ((@item-padding * 0.75) * 2);
-
-    .dir-rtl & {
-      margin-right: inherit;
-      margin-left: @icon-size + (@item-margin * 2) + ((@item-padding * 0.75) * 2);
-    }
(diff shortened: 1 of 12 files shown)


eleanor-heath (approved):

kirsty-hames (approved):

oliverfoster:
Oh, that's nice. Not ie11 compatible.

PR #488, opened 2024-01-23
New: Add drawer item locked state 

kirsty-hames:
Relates to Core issue to [allow for locked and inactive disabled buttons to be visually different](https://github.com/adaptlearning/adapt-contrib-core/issues/485).

`.is-locked` - applies to buttons that are temporarily disabled due to step locked content.

`.is-disabled` - applies to buttons with functionality not available.

Drawer item locked state inherits from disabled state by default so there will be no visual changes. By introducing `.is-locked` class, this gives the flexibility of setting locked styling that can differ from the standard disabled styling.

### DIFF
--- a/less/core/drawerItem.less
+++ b/less/core/drawerItem.less
@@ -13,6 +13,7 @@
     }
   }
 
+  &__item-btn.is-locked,
   &__item-btn.is-disabled {
     background-color: @disabled;
     color: @disabled-inverted;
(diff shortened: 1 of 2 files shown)


cahirodoherty-learningpool (approved):
👀

joe-allen-89 (approved):
👍

guywillis (approved):
👍

Issue #451, opened 2023-07-04
Replace hardcoded icon button border-radius with variable


kirsty-hames:
### Subject of the enhancement
Notify icon button `border-radius` is hardcoded at `50%`. This should be replaced with a variable as per `@item-border-radius` and `@btn-border-radius`.

Ref: issue raised on previous PR https://github.com/adaptlearning/adapt-contrib-vanilla/pull/448#discussion_r1235090469

Update: this also applies to [.drawer__btn](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/59e377d7a2da38ab29c596a278b7ce8374dfbe52/less/core/drawer.less#L14).

kirsty-hames:
This is being addressed in Adapt buttons #469.

PR #490, opened 2024-01-25
New: Adapt buttons overhaul (fixes #469)

### LINKED ISSUE #469: Adapt Buttons - Vanilla issue
Vanilla issue for core issue described here: https://github.com/adaptlearning/adapt-contrib-core/issues/386

### Your environment
* Master

guywillis:
Fixes: #469 

### New
* Expanded on [_colors.less variables](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/master/less/_defaults/_colors.less) based on the categories and states defined in ( https://github.com/adaptlearning/adapt-contrib-vanilla/issues/469#issuecomment-1910252410)
* Identified missing categories / states and updated as required.
* Created mixins to normalise button styles based on the catergories and states defined. Default button style set to 'filled' to mimic current Vanilla and 'outline' as optional style. The implementation allows for additional styles to be created for custom themes.
* Added UI btn related margins (nav, notify, drawer).
* Replace [hardcoded](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/62e4b3ef3ee6e6a26c9022b01dc15bc23ec7f661/less/core/drawer.less#L14) UI btn border-radius with [variable](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/2abcfad64fcac34d3e35915884bfc42cb12ec48a/less/_defaults/_spacing-config.less#L31) (fixes https://github.com/adaptlearning/adapt-contrib-vanilla/issues/451)
* Added focus states through out.
* Notify strapline simplification.
* Updated [properties.schema](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/b886dfec7b8ee10d51525c921d1db393bdfa6ac5/properties.schema#L7) and [theme.schema](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/b886dfec7b8ee10d51525c921d1db393bdfa6ac5/schema/theme.schema.json#L6) to reflect latest [_colors.less](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/issue/469/less/_defaults/_colors.less) variables.

### Note
The following has been implemented to prevent any breaking changes:
* Deprecated color variables left in for legacy support only (with comments). All new color variables inherit from existing color variables for backwards compatibilty.
* Deprecated `.notify-btn-icon` mixin left in for legacy support only (with comments). New `.ui-btn-ctrl` provided as fallback for backwards compatibilty.
* New color mixins applied to deprecated `.btn-text` and `.btn-icon` selectors for legacy support only (with comments).

### Summary
* For access to edit the new color variables in the AAT, [properties.schema](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/b886dfec7b8ee10d51525c921d1db393bdfa6ac5/properties.schema#L7) and [theme.schema](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/b886dfec7b8ee10d51525c921d1db393bdfa6ac5/schema/theme.schema.json#L6) have been updated to reflect the latest [_colors.less](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/issue/469/less/_defaults/_colors.less) variables.
* For access to edit the button styles in the AAT (e.g set an 'outline' button style), there's an outstanding PR awaiting work https://github.com/adaptlearning/adapt_framework/pull/3578.


StuartNicholls:
@guywillis, would it be a good idea to outline what is intended for Phase 2? Just thinking it might head off any unnecessary comments on the implantation?

properties.schema
```diff
@@ -111,21 +124,140 @@
             "type": "string",
             "inputType": "ColourPicker",
             "default": "",
-            "help": "Defines the Item text / icon colour when selected. Should be a colour that provides good contrast against the Item selected colour."
+            "help": "Defines the item text / icon colour when selected. Should be a colour that provides good contrast against the item selected colour."
           },
           "visited": {
             "title": "Visited colour",
```
kirsty-hames, line 130: As this variable is only used by presentation components I would change the title for consistency with the other component item colours.
_Item colour - visited_

properties.schema
```diff
@@ -111,21 +124,140 @@
...
             "type": "string",
             "inputType": "ColourPicker",
             "default": "",
-            "help": "Defines the Item colour when it has been visited."
+            "help": "Defines the item colour when it has been visited."
           },
           "visited-inverted": {
             "title": "Visited colour - inverted",
```
kirsty-hames, line 137: As this variable is only used by presentation components I would change the title for consistency with the other component item colours.
Item colour - inverted visited

properties.schema
kirsty-hames: _help_ text will need updating too.

properties.schema
kirsty-hames: _help_ text will need updating too.

properties.schema
```diff
@@ -111,21 +124,140 @@
...
+        "type": "object",
+        "required": false,
+        "title": "Content Items UI (action not content)",
+        "properties": {
+          "item-ui-color": {
+            "title": "Item UI colour",
+            "type": "string",
+            "inputType": "ColourPicker",
+            "default": "",
+            "help": "Defines the clickable areas in components that don't contain course content."
```
kirsty-hames, line 169: minor but 'selectable' is more inclusive word choice than 'clickable'

properties.schema
```diff
@@ -162,6 +294,58 @@
...
+          },
+          "btn-color-inverted-locked": {
+            "title": "Button colour - inverted locked",
+            "type": "string",
+            "inputType": "ColourPicker",
+            "default": "",
+            "help": "Defines the button text / icon colour when locked. Should be a colour that provides good contrast against the button locked colour."
+          },
+          "disabled": {
+            "title": "Disabled colour",
```
kirsty-hames, line 337: Update button title to _Button colour - locked_

schema/theme.schema.json
```diff
@@ -105,26 +120,140 @@
...
           "type": "string",
           "title": "Visited colour",
-          "description": "Defines the item colour when it has been visited",
+          "description": "Defines the item colour when it has been visited.",
           "default": "",
           "_backboneForms": "ColourPicker"
         },
         "visited-inverted": {
           "type": "string",
           "title": "Visited colour - inverted",
```
kirsty-hames, line 136: As this variable is only used by presentation components I would change the title for consistency with the other component item colours, _Item colour - inverted visited_.

schema/theme.schema.json
```diff
@@ -58,28 +59,29 @@
...
       "type": "object",
-      "title": "Items (components)",
+      "title": "Content Items (components)",
       "default": {},
       "properties": {
         "item-color": {
           "type": "string",
           "title": "Item colour",
-          "description": "Defines the clickable areas or points of interest in components",
+          "description": "Defines the clickable areas or points of interest in components.",
```
kirsty-hames, line 77: Minor but 'selectable' is more inclusive word choice than 'clickable'.

schema/theme.schema.json
```diff
@@ -105,26 +120,140 @@
...
+
+    "_itemsUi": {
+      "type": "object",
+      "title": "Content Items UI (action not content)",
+      "default": {},
+      "properties": {
+        "item-ui-color": {
+          "type": "string",
+          "title": "Item UI colour",
+          "description": "Defines the clickable areas in components that don't contain course content.",
```
kirsty-hames, line 165: Minor but 'selectable' is more inclusive word choice than 'clickable'.

schema/theme.schema.json
```diff
@@ -153,21 +282,73 @@
...
+        "btn-color-inverted-locked": {
+          "type": "string",
+          "title": "Button colour - inverted locked",
+          "description": "Defines the button text / icon colour when locked. Should be a colour that provides good contrast against the button locked colour.",
+          "default": "",
+          "_backboneForms": "ColourPicker"
+        },
+        "disabled": {
+          "type": "string",
+          "title": "Disabled colour",
```
kirsty-hames, line 330: Update button title to _Button colour - disabled_

schema/theme.schema.json
```diff
@@ -153,21 +282,73 @@
...
+          "default": "",
+          "_backboneForms": "ColourPicker"
+        },
+        "disabled-inverted": {
+          "type": "string",
+          "title": "Disabled colour - inverted",
```
kirsty-hames, line 336: Update button title to _Button colour - inverted disabled_

kirsty-hames (approved):
Thanks for updating schemas @guywillis. All looking good. I've just added a few text amends.

properties.schema
guywillis: This should now be resolved

properties.schema
guywillis: This should now be resolved

properties.schema
guywillis: This should now be resolved

properties.schema
guywillis: This should now be resolved

properties.schema
```diff
@@ -162,6 +294,58 @@
...
+            "type": "string",
+            "inputType": "ColourPicker",
+            "default": ""
+          },
+          "disabled-inverted": {
+            "title": "Disabled colour - inverted",
```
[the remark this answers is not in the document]
guywillis: This should now be resolved

schema/theme.schema.json
```diff
@@ -105,26 +120,140 @@
...
           "type": "string",
           "title": "Item colour - inverted selected",
-          "description": "Defines the item text/icon colour when selected. Should be a colour that provides good contrast against the Item selected colour",
+          "description": "Defines the item text / icon colour when selected. Should be a colour that provides good contrast against the item selected colour.",
           "default": "",
           "_backboneForms": "ColourPicker"
         },
         "visited": {
           "type": "string",
           "title": "Visited colour",
```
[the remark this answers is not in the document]
guywillis: This should now be resolved

schema/theme.schema.json
guywillis: This should now be resolved

schema/theme.schema.json
guywillis: This should now be resolved

schema/theme.schema.json
guywillis: This should now be resolved

schema/theme.schema.json
guywillis: This should now be resolved

kirsty-hames (approved):

swashbuck:
Hi, @guywillis . This is looking great. Thanks for the hard work. I have a few comments/questions below.

1. For disabled outline style buttons, I think I would prefer just a lighter border instead of a filled background. I do like the filled style on hover/focus, though.

Outline

2. In the Less filenames, I think the 'i' should be capitalized in 'UI' since it is an acronym. Ex: __drawerUIStyles.less_ not __drawerUiStyles.less_. Microsoft [says](https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71)) "When using acronyms, use Pascal case or camel case for acronyms more than two characters long. For example, use HtmlButton or htmlButton. However, you should capitalize acronyms that consist of **only two characters**, such as System.IO instead of System.Io." I'm sure you could find sources that disagree, of course.
3. _core/questionButton.less_ - "button" should come first like _core/buttonQuestion.less_. This is similar to the notify/notifyPush and drawer/drawerItem Less files and would keep these button files together.
4. `btn-color-outline()`: Why use a box shadow and not an actual border? I get why we should not use outline since that should be reserved for focus states.
5. This may be out of scope, but I do not like that some variable names include "color" (ex. @menu-item-btn-color-hover) while others do not (ex. @progress-border). Without including "color" or something similar (ex. "bg-color"), it's unclear what the variable is for. For instance, @progress-border could define the border width or the entire border shorthand style.

Otherwise, great work!

guywillis:
Hey @swashbuck, thank you for your comments and feedback.

> 1. For disabled outline style buttons, I think I would prefer just a lighter border instead of a filled background. I do like the filled style on hover/focus, though.

I agree, I think I would prefer this visual too. Our initial reasoning for keeping the filled disabled buttons for the outline style were two fold:

1. It provides a very clear visual indication that the button is disabled, especially in contrast to an active outline style button.
2. It provides simplicity in the Vanilla theme Less.

I would not be adverse to changing this but having experimented with mimicking this change in others button areas it looks weaker to me than the filled version. 

**Item**
| Filled | Outline | 
| --- | --- | 
| item-filled | item-outline | 

**Item UI**
| Filled | Outline |
| --- | --- | 
| itemUi-filled | itemUi-outline |

**Notify UI**
| Filled | Outline |
| --- | --- | 
| notifyUi-filled | notifyUi-outline |

> 2. In the Less filenames, I think the 'i' should be capitalized in 'UI' since it is an acronym. Ex: __drawerUIStyles.less_ not __drawerUiStyles.less_. Microsoft [says](https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71)) "When using acronyms, use Pascal case or camel case for acronyms more than two characters long. For example, use HtmlButton or htmlButton. However, you should capitalize acronyms that consist of **only two characters**, such as System.IO instead of System.Io." I'm sure you could find sources that disagree, of course.

I am not against such a change, it makes sense. 

Reflecting upon the names of the files `_itemStyles.less`, `_btnStyles.less` etc, I think we can actually drop the `Styles` part and instead rename the folder from `_buttonMixins` to `_buttonStyles`

> 3. _core/questionButton.less_ - "button" should come first like _core/buttonQuestion.less_. This is similar to the notify/notifyPush and drawer/drawerItem Less files and would keep these button files together.

This change makes a lot of sense, I am in favour. 

> 4. `btn-color-outline()`: Why use a box shadow and not an actual border? I get why we should not use outline since that should be reserved for focus states.

Predominantly, I prefer box shadow over border as it's more flexible:

* box shadow doesn't affect the elements width or height
* it can be applied outside the dimensions of the element as well as the inside
* it can support multiple values at once

> 5. This may be out of scope, but I do not like that some variable names include "color" (ex. @menu-item-btn-color-hover) while others do not (ex. @progress-border). Without including "color" or something similar (ex. "bg-color"), it's unclear what the variable is for. For instance, @progress-border could define the border width or the entire border shorthand style.

I am in agreement here. It should be consistent and flow together. I think it would be best if we addressed this as a separate issue.

less/_defaults/_spacing-config.less
```diff
@@ -2,31 +2,75 @@
...
 @icon-padding: 1rem;
 
-// Item
+// Content Item
 // --------------------------------------------------
 @item-margin: 0.5rem;
 @item-padding: 1rem;
 
 @item-border-radius: 0.25rem;
+@item-color-border-width: 1px;
```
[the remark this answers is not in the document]
guywillis: We opted for using px instead of rem here as px is consistent regardless of browser zoom and font size.

swashbuck:
Thanks, @guywillis . I agree with all your points. Fine with leaving the disabled outline styles as is. Should be simple enough to change in a custom theme.

kirsty-hames:
Hi @guywillis / @swashbuck, I've had a read through all comments [above](https://github.com/adaptlearning/adapt-contrib-vanilla/pull/490#issuecomment-2598219854) and I'm in agreement. I just have one preference regarding the following, 

> Reflecting upon the names of the files `_itemStyles.less`, `_btnStyles.less` etc, I think we can actually drop the `Styles` part and instead rename the folder from `_buttonMixins` to `_buttonStyles`

I agree with dropping 'Styles' from the files but think it makes sense to keep the folder as `_buttonMixins`. As these files contain mixins only. 'Styles' is always a given with CSS anyway :)

swashbuck (approved):
👍

oliverfoster:
We need one more +1 here from a non-kineo. 3 weeks outstanding.

PR #494, opened 2024-02-08
Fix: Change @max-width to 100%, remove max-width: inherit from .page (fixes #493)

### LINKED ISSUE #493: @max-page-width variable is ignored for .page element
### Subject of the issue
The max-width of `.page` should be using the `@max-page-width` [value](https://github.com/adaptlearning/adapt-contrib-core/blob/0575f81b7cfb59f95fdb120e1b4aa58d3afd1397/less/core/page.less#L2). Instead, the Vanilla theme is [overriding](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/6d6946dee68141fe37e5072c7d6fde9b38965fb2/less/core/page.less#L3) this with a `max-width: inherit` style.

Is this intentional? If not, should we remove the `max-width: inherit` style?

max-width-page

### Your environment
* Vanilla 9.14.1
* FW 5.35.0

### To reproduce
In your theme's CSS:
```
@body-background-color: #96D5E4;

.page {
  background-color: @white;
}
```

### Expected results
Blue body background color is visible on the sides.

### Actual results
Entire page is white because `.page` is the full window width and does not use `@max-page-width`.

swashbuck:
Fixes #493 

### Fix
* Changes `@max-width` to 100%
* Removes `max-width: inherit` from the `.page` element

### Testing
1. In your theme's CSS:
```
@body-background-color: #96D5E4;

.page {
  background-color: @white;
}
```
2. Change `@max-width` variable to 90rem

### Expected results
Blue body background color is visible on the sides.

### DIFF
--- a/less/_defaults/_spacing-containers.less
+++ b/less/_defaults/_spacing-containers.less
@@ -5,8 +5,7 @@
 @device-width-large: unit(@adapt-device-large, rem);
 @device-width-xlarge: unit(@adapt-device-xlarge, rem);
 
-@max-width: 90rem;
-
+@max-width: 100%;
 @max-content-width: @device-width-large;
 @max-page-width: @max-width;
 
--- a/less/core/page.less
+++ b/less/core/page.less
@@ -1,6 +1,5 @@
 .page {
   position: relative;
-  max-width: inherit;
   .l-container-padding(@page-padding-ver, @page-padding-hoz);
 
   &.has-bg-image > .background {


cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

joe-allen-89 (approved):
👍  working as expected.

PR #495, opened 2024-02-08
Fix: Utilize .size-small class for .hide-on-mobile class (fixes #492)

### LINKED ISSUE #492: hide-on-mobile class breakpoint issue
### Subject of the issue
The `.hide-on-mobile` class allows you to hide a component for mobile views. The problem is that it's not properly aligned with the layout breakpoints. This can cause a component with `.hide-on-mobile` to disappear at the `@device-width-medium` breakpoint.

**hide-on-mobile**
[src/theme/adapt-contrib-vanilla/less/project/theme-common.less](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/6d6946dee68141fe37e5072c7d6fde9b38965fb2/less/project/theme-common.less#L95)

### Steps to reproduce
1. Add a single-width Graphic and Text component to the same block.
2. For the Graphic component, add the `hide-on-mobile` class.
3. View the page exactly at the medium breakpoint (720px width by default).

### Expected behaviour
Graphic component should still be visible since 720px is "medium" and higher.

### Actual behaviour
- Text component correctly becomes 50% width.
- Graphic component is incorrectly hidden. Graphic component reappears at 721px and higher.

### Solution 
I think the breakpoint should be adjusted by 1px (0.0625rem).

```
.hide-on-mobile {
  @media (max-width: (@device-width-medium - 0.0625rem)) {
    display: none;
  }
}
```

swashbuck:
Fix #492 

### Fix
* Utilize the `.size-small` html class to determine when `.hide-on-mobile` applies

### Testing
See #492

### DIFF
--- a/less/project/theme-common.less
+++ b/less/project/theme-common.less
@@ -92,8 +92,6 @@
   }
 }
 
-.hide-on-mobile {
-  @media (max-width: @device-width-medium) {
-    display: none;
-  }
+.size-small .hide-on-mobile {
+  display: none;
 }


guywillis (approved):
👀

cahirodoherty-learningpool (approved):
👀

oliverfoster (approved):

PR #503, opened 2024-03-13
Fix: Optimise `blend.less` (fixes #502)

### LINKED ISSUE #502: Optimise `blend.less`
### Subject of the issue
The Less loop found in `blend.less` file is a bit of a sledge hammer to crack a nut and could be simpler and more efficient.

We should be able to utilise the `each` list function to extract the necessary values from the `@blend-modes` variable and assign the value to the associated property.

https://lesscss.org/functions/#list-functions-each

### Your environment
* Master FW & AT

guywillis:
Fixes: #502 

### Fix
* Utilise the each list function to optimise the `blend.less` file

### DIFF
--- a/less/_defaults/blend.less
+++ b/less/_defaults/blend.less
@@ -5,16 +5,12 @@
 // --------------------------------------------------
 @blend-modes: color, color-burn, color-dodge, darken, difference, exclusion, hard-light, hue, lighten, luminosity, multiply, overlay, saturation, screen, soft-light;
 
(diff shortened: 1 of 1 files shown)


kirsty-hames (approved):
Works as expected thanks - testing on Mac Safari, Chrome and FF.

cahirodoherty-learningpool (approved):
👀

swashbuck (approved):
👍

PR #505, opened 2024-03-28
New: Add margin start and margin end column classes (fixes #504)

### LINKED ISSUE #504: Column push / pull classes
Now that we have [columns](https://github.com/adaptlearning/adapt-contrib-vanilla/wiki/Custom-Classes#columns), it would be useful to also have push and pull classes like with [Bootstrap](https://getbootstrap.com/docs/3.4/css/#grid-column-ordering). These would be used to indent components by a certain number of columns from the left and/or right.

For example, `col-push-md-1` would left indent a component by 1 column width for medium breakpoints and higher. For RTL, it would indent from the right instead.

Pull classes would work the other way. `col-pull-md-1` would right indent a component by 1 column width for medium breakpoints and higher. For RTL, it would indent from the left instead.

I had a need for these classes in a recent project (see below), so I thought it may be useful to include in the Vanilla theme.

swashbuck:
Fixes #504 

### New
* Adds new margin start and margin end column classes (e.g. `margin-start-sm-2` and `margin-end-md-1`) for indenting components from the left or right.

### Testing
Configure two single width components using the existing [column](https://github.com/adaptlearning/adapt-contrib-vanilla/wiki/Custom-Classes#columns) classes and the new margin classes. For example, in the same block:

**Single width Text component**
Add classes `col-md-5 margin-start-md-1`. This makes the component 5 columns wide with a 1 column margin from the left at the medium breakpoint and higher.

**Single width Graphic component**
Add classes `col-md-4 margin-end-md-1`.  This makes the component 4 columns wide with a 1 column margin from the right at the medium breakpoint and higher.

The result for the above would look something like the following (gray block background color and guidelines added for illustrative purposes):

margin-start-end

The number of columns + starting/ending margins must add up to the total number of columns (or less) which is 12 unless redefined in the theme.

### Additional work
The theme [wiki](https://github.com/adaptlearning/adapt-contrib-vanilla/wiki/Custom-Classes#columns) should be updated with these new classes.

### DIFF
--- a/less/project/columns.less
+++ b/less/project/columns.less
@@ -1,3 +1,8 @@
+// Custom classes
+// e.g. 'col-sm-8 col-md-10'
+// min small viewport takes up 8 out of 12 columns
+// min medium viewport takes up 10 columns
+// if undefined element defaults to 100% width
 .make-columns(@size, @media-query, @count, @iteration: 1) when (@count >= @iteration) {
   @media (min-width: @media-query) {
     .boxmenu-item.col-@{size}-@{iteration} {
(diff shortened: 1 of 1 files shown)


less/project/columns.less
```diff
@@ -1,3 +1,8 @@
+// Custom classes
```
guywillis, line 1: Cheeky of me, can we change to `Custom column classes`

less/project/columns.less
```diff
@@ -11,13 +16,56 @@
...
+    }
+    .component.col-push-@{size}-@{iteration} {
+      margin-inline-start: (@iteration * 100% / @count);
+    }
+  }
+
+  .make-columns-push(@size, @media-query, @count, (@iteration + 1));
+}
+
+.make-columns-push(xs, 0, 12); // deprecated: backward compatibility only
```
guywillis, line 43: Don't believe this is required for new functionality

less/project/columns.less
```diff
@@ -11,13 +16,56 @@
...
+    }
+    .component.col-pull-@{size}-@{iteration} {
+      margin-inline-end: (@iteration * 100% / @count);
+    }
+  }
+
+  .make-columns-pull(@size, @media-query, @count, (@iteration + 1));
+}
+
+.make-columns-pull(xs, 0, 12); // deprecated: backward compatibility only
```
guywillis, line 67: Don't believe this is required for new functionality

guywillis:
Lovely work @swashbuck 👍

less/project/columns.less
swashbuck: Yep, that's more descriptive. Done.

I've also updated the push and pull comments (e.g. "Custom column pull classes" instead of "Custom pull classes"), but happy to change them to whatever.

guywillis (approved):
👍

oliverfoster:
Why calling them pull and push and not left and right?

swashbuck:
> Why calling them pull and push and not margin left and right?

@oliverfoster Since these classes can be used with RTL layouts, "left" and "right" would not work. As for "push" and "pull", these were the names that Bootstrap used for the same sort of utility class. So, my thought was that they would be familiar to developers who have used Bootstrap. However, they may be deprecated in the newest version(s).

> I'd agree. Perhaps margin start and margin end to align with the CSS?

@kirsty-hames Yep, I'm fine with this if everyone else is? "Start" and "end" would make sense with both RTL and LTR. Something like this then?

`margin-start-sm-2`
`margin-end-md-1`

swashbuck:
Yep, deprecated as of v4:
https://getbootstrap.com/docs/4.0/migration/#grid-system-1

> Dropped push and pull modifier classes for the new flexbox-powered order classes. For example, instead of .col-8.push-4 and .col-4.pull-8, you’d use .col-8.order-2 and .col-4.order-1.

joe-allen-89 (approved):
👍

swashbuck:
### To do:

- Change class names to include "margin". For example: `margin-start-sm-2` and `margin-end-md-1`
- Change CSS classes/mixins. `.make-columns-pull .make-columns-push`
- Update documentation and examples where necessary

swashbuck:
Class names have been updated. Ready for re-review.

kirsty-hames (approved):

oliverfoster (approved):

cahirodoherty-learningpool (approved):
👀

PR #540, opened 2024-12-18
Fix: Add article to bg-color mixin (fix #539)

### LINKED ISSUE #539: Include articles in bg-color mixin
Articles should work with the `bg-color` mixin. Currently, the content in an article header is not included. Only blocks and components are targeted.

https://github.com/adaptlearning/adapt-contrib-vanilla/blob/b886dfec7b8ee10d51525c921d1db393bdfa6ac5/less/project/theme-common.less#L18-L27

It is also mentioned that the class can be added to articles, but this is not true.
https://github.com/adaptlearning/adapt-contrib-vanilla/blob/b886dfec7b8ee10d51525c921d1db393bdfa6ac5/less/project/theme-common.less#L6

There is an ongoing discussion in #304 . For now, I think we should just include articles in the existing mixin.

### Steps to reproduce
1. Add a `bg-color` class to an article
2. Add displayTitle, body, and instruction to the article

### Expected behaviour
Text colors should be changed in the article header text

swashbuck:
Fixes #539 

### Fix
* Adds article to the `bg-color` mixin

### Testing
1. Add a `bg-color` class to an article
2. Add `displayTitle`, `body`, and `instruction` values to the article

### Expected behaviour
Text colors should be changed in the article header text

### DIFF
--- a/less/project/theme-common.less
+++ b/less/project/theme-common.less
@@ -19,6 +19,7 @@
   background-color: @@color;
   color: @@color-inverted;
 
+  .article,
   .block,
   .component {
     &__title,


guywillis:
Hey @swashbuck,

Adding `.article` to the mixin will allow background colour application to the article but the associated title / body etc styles won't apply to the child blocks and articles, necessitating further classes on lower levels.

Is this behaviour as expected?

swashbuck:
> Adding `.article` to the mixin will allow background colour application to the article but the associated title / body etc styles won't apply to the child blocks and articles, necessitating further classes on lower levels.

@guywillis Actually, I think it would, but you can still override it on child elements.

For instance, here I have a `bg-color` only on the article. The component inherits the colors.


If I add a `bg-color` to the component, then I can override the colors:


Is this an unintended side effect?

guywillis:
I believe I have made a mistake and misunderstood what impact. 

It appears, and confirming through testing, as though adding an article class to the mixin works as required. 

Please disregard the comments in my previous message.

cahirodoherty-learningpool (approved):
👀

kirsty-hames (approved):
👍

PR #548, opened 2025-06-09
Update: Apply @progress-border-radius to PLP indicator inner (fixes #547)

### LINKED ISSUE #547: pageLevelProgress indicator missing border-radius override
### Subject of the issue
You can set the border radius of the indicator via `@progress-border-radius` however this is only applied to the [outer indicator border](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/c03b5484783f7b42ed178d3aa39e5d552d4e2648/less/plugins/adapt-contrib-pageLevelProgress/pageLevelProgressIndicator.less#L12C20-L12C44) not the [inner](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/c03b5484783f7b42ed178d3aa39e5d552d4e2648/less/plugins/adapt-contrib-pageLevelProgress/pageLevelProgressIndicator.less#L16). So the initial plugin [50px value](https://github.com/adaptlearning/adapt-contrib-pageLevelProgress/blob/d61a65e097143dfeab014c240f7c6a5f8689dcf6/less/pageLevelProgressIndicator.less#L19) remains.

Default 50px value:
Image

Change `@progress-border-radius` to `0`:
Image

### Expected behaviour
Apply `@progress-border-radius` to the [inner](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/c03b5484783f7b42ed178d3aa39e5d552d4e2648/less/plugins/adapt-contrib-pageLevelProgress/pageLevelProgressIndicator.less#L16):

Image

kirsty-hames:
Fixes https://github.com/adaptlearning/adapt-contrib-vanilla/issues/547

### Update
* Apply `@progress-border-radius` to PLP indicator inner. 

### Testing
1. Change `@progress-border-radius` to `0`.
2. Review PLP indicators in your course. The curve should be removed from the indicator outer border and the inner bar.

Default 50px value:
Image

Changing `@progress-border-radius` to `0` should look like this:
Image

Instead of this:
Image

### DIFF
--- a/less/plugins/adapt-contrib-pageLevelProgress/pageLevelProgressIndicator.less
+++ b/less/plugins/adapt-contrib-pageLevelProgress/pageLevelProgressIndicator.less
@@ -14,6 +14,7 @@
   }
 
   &__indicator-inner {
+    border-radius: @progress-border-radius;
     background-color: @progress-inverted;
   }
 


cahirodoherty-learningpool (approved):
👀

guywillis (approved):
👀

PR #552, opened 2025-06-18
Fix: Reduced margin-start / end iteration default value (fixes #551)

### LINKED ISSUE #551: Unable to override column margin responsively
### Subject of the issue
Bug encountered when using `margin-start-XX-YY` and `margin-end-XX-YY` custom classes to set the start and end margin for components. 

If a user were to set a margin on a smaller screen size and attempt to remove it entirely at a larger resolution they would be unable to do so.

e.g.

```
"_classes": "margin-start-md-4 margin-start-lg-0",
```

### Your environment
* Master AT and FW

### Expected behaviour
User should be able to remove the start margin using a value of 0.

guywillis:
Fixes #551 

### Fix
* Reduced margin-start / end mixin iteration default value to enable overrides using 0 value

### DIFF
--- a/less/project/columns.less
+++ b/less/project/columns.less
@@ -26,7 +26,7 @@
 // e.g. 'margin-start-sm-2 margin-start-md-1'
 // min small viewport indents 2 columns from the left (or right for RTL)
 // min medium viewport indents 1 column from the left (or right for RTL)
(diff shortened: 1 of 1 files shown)


kirsty-hames (approved):
👍

PR #559, opened 2025-07-17
Fix: Add end margin only when question is complete (fix #558)

### LINKED ISSUE #558: Submit button alignment issue
### Subject of the issue

The action/submit button appears to have a margin even when the feedback button is not present. This can lead to the button being slightly off-centered.

Same issue as:
https://github.com/adaptlearning/adapt-contrib-core/issues/650

swashbuck:
Fix #558 

### Fix
* Adds end margin only when question is complete

### Related
* Needs this Core PR approved as well: https://github.com/adaptlearning/adapt-contrib-core/pull/683

### DIFF
--- a/less/core/buttonQuestion.less
+++ b/less/core/buttonQuestion.less
@@ -39,8 +39,8 @@
     border-radius: 50%;
   }
 
-  .can-show-marking &__action.is-full-width,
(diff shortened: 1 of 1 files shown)


kirsty-hames (approved):

cahirodoherty-learningpool (approved):
👀

PR #581, opened 2026-02-25
New: Add page header color variables (fix #579)

swashbuck:
Fix #579 

### Fix
* Add new page header background and text color variables for easier styling
* No new authoring tool settings were added.

### DIFF
--- a/less/_defaults/_colors.less
+++ b/less/_defaults/_colors.less
@@ -136,6 +136,14 @@
 @progress-inverted: @white;
 @progress-border: @greyDark;
 
+// Page
+// --------------------------------------------------
+@page-header-background-color: transparent;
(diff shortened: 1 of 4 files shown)


kirsty-hames (approved):
Thanks @swashbuck. I've previously used the [header color mixin](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/998fa96ddd90f20a1883c52a14d3b891c352d788/less/project/theme-contentObjects.less#L19) for applying page header colors but makes sense to have global color variables then use the mixin for custom colors per page (if needed). 

Will you be adding the new variables to the theme schemas as part of this PR?

kirsty-hames (approved):

PR #585, opened 2026-03-10
New: Extend column layout support to blocks (fixes #584)

kirsty-hames:
Fixes https://github.com/adaptlearning/adapt-contrib-vanilla/issues/584

### New
* Extended the existing responsive 12-column layout support (width and margin-start/margin-end mixins) to blocks, bringing them in line with the existing column layout behaviour for components and Boxmenu items.
* Added optional `.contain-content-container` article class, which constrains the child `block__container` width to `@max-content-width` (70rem/1120px) when using block column layout (keeping inline with the default content width). Without this, the block container spans the full course width giving greater flexibility in layout design.
* **Responsive column layout** link added to [Vanilla wiki home](https://github.com/adaptlearning/adapt-contrib-vanilla/wiki) so it's not just buried under **Custom Classes**.
* [Columns wiki](https://github.com/adaptlearning/adapt-contrib-vanilla/wiki/Custom-Classes#columns) updated to include block support and AAT considerations. **Note, release versions will need adding once PR has been merged.**

### Example
The screenshot below demonstrates two blocks within a single article, each set to 50% width via `col-lg-6`. Below the large breakpoint, both blocks stack to full width.
- **Left block** - contains a Text component (`"_layout": "full"`)
- **Right block** - contains a Graphic and an MCQ component (both `"_layout": "full"`)
Image

### Testing
1. Add a custom class of `col-<size>-<n>` (e.g. `col-md-6`) to a block and verify the block renders at the expected column width at the target breakpoint.
3. Confirm the block's container switches to a flex row layout at or above the medium breakpoint (`@device-width-medium`).
4. Test `margin-start-<size>-<n>` and `margin-end-<size>-<n>` custom classes on a block and verify the correct inline margins are applied.
5. Add the `.contain-content-container` class to an article containing blocks with column layout classes and verify the `block__container` is constrained to the expected maximum content width.
6. Verify that existing column layout behaviour for components and Boxmenu items is unaffected.


less/project/theme-article.less
```diff
@@ -0,0 +1,10 @@
+// --------------------------------------------------
+// Article
+// --------------------------------------------------
+
+// Contain width of child block to 70rem (1120px)
+// Optional class to contain content width when using block column layout
+// --------------------------------------------------
+.article.contain-content-container .block__container {
```
oliverfoster: Does this need to have the `:has` directive as well?

kirsty-hames: As it's an optional class I didn't think it was necessary, however, there's no use case for this class unless you're using block columns. Applying the `:has` directive will prevent additional styles being applied to `.block__container` at screen sizes where columns aren't set. So in short, yes I'll add this thanks :)

oliverfoster (approved):

swashbuck (approved):
👍

swashbuck (approved):
👍

PR #596, opened 2026-04-06
New: Expand header-color mixin to articles and blocks, add prefixed colors (fix #595)

### LINKED ISSUE #595: Expand header-color mixin to articles and blocks

The current `.header-color-mixin` allows us to add colors to the menu and page header. Now that we can add background images to articles and blocks (https://github.com/adaptlearning/adapt-contrib-vanilla/issues/571), it would be useful to expand this mixin to those elements as well.

https://github.com/adaptlearning/adapt-contrib-vanilla/blob/c9ac1c15c9a365f9ccd8200bdaf94e2f167e65e8/less/project/theme-contentObjects.less#L19-L52

swashbuck:
Fix #595 

### New
* Expand `header-color mixin` to article and block headers
* Add prefixed color class options to `header-color` and `bg-color` mixins. This allows both classes to be used without conflict on a single article or block.

### Testing
1. **Header color:** Add `header-color black` to an article or block with header text. The header background should be black with white text. 

2. **Header color + element color:** Add `header-color header-black` and `bg-color blue` to an article or block with header text. The header background should be black with white text while the rest of the article/block should be blue with white text.  Note that you will need to create a bg-color class for blue with `.bg-color-mixin(blue, white);`

### DIFF
--- a/less/project/theme-common.less
+++ b/less/project/theme-common.less
@@ -81,6 +81,45 @@
 
 }
 
+// Header color mixin
+// Add to menu/page/article/block to enable
+// e.g. 'header-color transparent-dark'
+// Note: both arguments must be predefined variables
+// --------------------------------------------------
+.header-color-mixin(black, white);
+.header-color-mixin(background, background-inverted);
+.header-color-mixin(transparent-light, font-color);
+.header-color-mixin(transparent-dark, font-color-inverted);
+
+.header-color-mixin(@color, @color-inverted) {
+
+.header-color.@{color} {
+  .menu,
+  .page,
+  .article,
+  .block {
+    &__header {
+      background-color: @@color;
+    }
+  }
+
+  .menu,
+  .page,
+  .article,
+  .block {
+    &__title,
+    &__subtitle,
+    &__body,
+    &__body a,
+    &__instruction,
+    &__instruction a {
+      color: @@color-inverted;
+    }
+  }
+}
+
+}
+
 // styles associated with _textAlignment property
 // --------------------------------------------------
 .text-align-mixin(page);
--- a/less/project/theme-common.less
+++ b/less/project/theme-common.less
@@ -4,7 +4,7 @@
 
 // Background color mixin
 // Can be added to course, contentObject, article, block, or component to enable
-// e.g. 'bg-color black'
+// e.g. 'bg-color black' or 'bg-color bg-black' (prefixed, safe to combine with header-color)
 // Note: both arguments must be predefined variables
 // --------------------------------------------------
 .bg-color-mixin(black, white);
@@ -15,7 +15,8 @@
 
 .bg-color-mixin(@color, @color-inverted) {
 
-.bg-color.@{color} {
+.bg-color.@{color},
+.bg-color.bg-@{color} {
   background-color: @@color;
   color: @@color-inverted;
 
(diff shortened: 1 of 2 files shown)


kirsty-hames (commented):
Hey @swashbuck, thanks for the PR. I've tested this and have a few observations:

**1. Inherited styles on child headers**

Applying the `header-color` mixin at page or article level cascades styles down to all child headers, not just the targeted model. For example, applying the mixin class to an article causes the child block header to inherit the same styles (see screenshot). The same behaviour is observed when applied at page level. I'm assuming we'd want to tighten the scope to style the direct header only?

Image


**2. Completion indicator (PLP) styles missing**

Since completion indicators can be applied at both article and block level, the mixin should account for this. Suggest including something like the following:

```less
.is-article,
.is-block {
  .pagelevelprogress {
    &__indicator {
      border-color: @@color-inverted;
    }
    &__indicator-inner {
      background-color: @@color;
    }
    &__indicator-bar {
      background-color: @@color-inverted;
    }
  }
}
```
PLP not styled by mixin:
Image

PLP styled by mixin:
styled-plp



**3. Header padding**

Article and block headers currently have no default padding, so the background colour sits flush against the text. The existing `.comp-header-bg-color` mixin, [handles the padding as part of its implementation](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/c9ac1c15c9a365f9ccd8200bdaf94e2f167e65e8/less/project/theme-components.less#L23). Should we follow the same pattern here, or is the expectation that padding will be added via custom theme CSS?

Image

swashbuck:
Thanks, @kirsty-hames 

**1. Inherited styles on child headers** - Should now be fixed.

**2. Completion indicator (PLP) styles missing** - Implemented. For simplicity, I'm applying to menu and page as well even though they don't currently allow completion indicators.  Might be something that gets added in the future. Otherwise, do you think we should rewrite to only apply to articles and blocks? It would make it slightly more complicated but doable.

**3. Header padding** - Still thinking about this. We already have [block header](https://github.com/adaptlearning/adapt-contrib-vanilla/blob/c9ac1c15c9a365f9ccd8200bdaf94e2f167e65e8/less/_defaults/_spacing-containers.less#L52) spacing variables. However, they default to `0` and we probably don't want to have the same padding for block headers with _and_ without background colors. Separately, I was going to propose a new `.expand-header` class as well that would make the block headers full width.

**4. Conflict with bg-color (new concern)** - I've realized that we would not be able to use both header-color and bg-color since they both rely on a non-prefixed color class. For instance, we couldn't set the block header background color to black and the block background color to blue due to how the mixins are set up.

`header-color black bg-color blue`

One solution could be to allow colors to be prefixed so that it's clear which mixin it applies to. This would be backwards compatible since it would only introduce an alternate way to specify a color.

`header-color header-black bg-color bg-blue`

bg-color mixin
```
.bg-color-mixin(@color, @color-inverted) {
  .bg-color.@{color},
  .bg-color.bg-@{color} {
    // ... same body as today
  }
}
```

header-color mixin
```
.header-color-mixin(@color, @color-inverted) {
  .header-color.@{color},
  .header-color.header-@{color} {
    // ... same body as today
  }
}
```

Thoughts on this or any other ideas?

swashbuck:
@kirsty-hames I've implemented issue 4 in https://github.com/adaptlearning/adapt-contrib-vanilla/pull/596/commits/ca7d5e09ba4cad27cde9acb2802bb9e519771889 , but let me know what you think.

kirsty-hames:
> @kirsty-hames I've implemented issue 4 in [ca7d5e0](https://github.com/adaptlearning/adapt-contrib-vanilla/commit/ca7d5e09ba4cad27cde9acb2802bb9e519771889) , but let me know what you think.

This works for me thanks @swashbuck. Issue resolved and it still has legacy support for existing courses.

lemmyadams (approved):

kirsty-hames (approved):

PR #598, opened 2026-04-21
Fix: Adjust block container only when containing a block with a column class (fix #597)

### LINKED ISSUE #597: Incorrect block__container when using column classes on component
### Subject of the issue
When using column classes on a component, the `block__container` element is incorrectly sized. The container should only be affected when there are _blocks_ that contain column classes.  Caused by #585 .

### Expected result
The block continues to be full width.

Image

### Actual result
The block width is constrained.

Image

swashbuck:
Fix #597 

### Fix
* Adjust block container only when containing a _block_ with a column class. Adding a column class to a _component_ alone should not affect the block container width.

### Testing (use case 1)
Test using column classes on components.

1. Create a block with a single width Text and Graphic component.
2. Add `bg-color black` to the block
3. Add `col-md-4` to the Graphic component

### Expected results
The black block background should continue to span 100% of the page width.

fw

### Testing (use case 2)
Test that the new block column classes still work as expected.

1. Add `col-md-8` to a block
2. Add `bg-color black` to the block

### Expected results
Block container should be content width rather than full page width. The actual block itself should be 8 columns wide.

block

### DIFF
--- a/less/project/columns.less
+++ b/less/project/columns.less
@@ -17,7 +17,7 @@
 
     // Additional styles required to extend column layout support to blocks
     // for consistency with component column styles in core/less/core/component.less
-    .block__container:has(.col-@{size}-@{iteration}) {
+    .block__container:has(.block.col-@{size}-@{iteration}) {
       width: 100%;
       display: flex;
       flex-direction: column;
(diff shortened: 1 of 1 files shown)


kirsty-hames (approved):
Good spot @swashbuck! Apologies I missed this with the initial PR. This now works as expected 👍

lemmyadams (approved):

